PR #4, opened 2020-07-15
Dependency mappings

### LINKED ISSUE #3: Support a platform-dependent dependency mapping
Currently, online packaged buildpacks must download their contributed dependency from the open internet.  There are situations where the URIs for those download locations are blocked from within a corporate network.  `libpack`'s `DependencyCache` should noticed the existence of an artifact or artifacts in `/platform` that defines a mapping between a buildpack's dependency (id/version) and a different URL that it should be downloaded from. This should happen transparently (save for a different log message) to both the buildpack author using `DependencyCache` and the application developer using a buildpack that uses `DependencyCache`.

ekcasey:
Resolves #3


effect/mocks/executor.go
```diff
@@ -3,8 +3,9 @@
 package mocks
 
 import (
-	effect "github.com/paketo-buildpacks/libpak/effect"
```
nebhale, line 6: This is a generated file that's going to be reverted each time we regenerate the file.  Probably shouldn't update it as part of this PR.

dependency_mapping.go
```diff
@@ -0,0 +1,54 @@
...
+package libpak
+
+import (
+	"path/filepath"
+
+	"github.com/BurntSushi/toml"
+)
+
+// MappingsFile defines dependency mappings for a set of buildpacks
+type MappingsFile struct {
```
nebhale, line 26: Why not hide this implementation details?

dependency_mapping.go
```diff
@@ -0,0 +1,54 @@
...
+
+// DependencyMapping defines a new URI for a dependency with the given ID and Version
+type DependencyMapping struct {
+	ID      string `toml:"id"`
+	Version string `toml:"version"`
+	URI     string `toml:"uri"`
+}
+
+// ReadMappingsFile read MappingsFile from path
+func ReadMappingsFile(path string) (MappingsFile, error) {
```
nebhale, line 43: By returning `([]BuildpackMappings, error)` instead?

dependency_mapping.go
```diff
@@ -0,0 +1,54 @@
...
+	mappingsFile := MappingsFile{}
```
nebhale, line 44: And using an anonymous struct

```go
raw := struct{
  BuildpackMappings []BuildpackMappings `toml:"buildpacks"`
}{}
```

dependency_cache.go
```diff
@@ -48,16 +48,31 @@ type DependencyCache struct {
...
 }
 
 // NewDependencyCache creates a new instance setting the default cache path (<BUILDPACK_PATH>/dependencies) and user
 // agent (<BUILDPACK_ID>/<BUILDPACK_VERSION>).
-func NewDependencyCache(buildpack libcnb.Buildpack) DependencyCache {
+// Mappings will be read from <PLATFORM_DIR>/dependencies/mappings.toml
+func NewDependencyCache(context libcnb.BuildContext) (DependencyCache, error) {
+	mappingsFile, err := ReadMappingsFile(DefaultMappingsFilePath(context.Platform.Path))
+	if err != nil {
+		return DependencyCache{}, err
```
nebhale, line 62: More descriptive error: `fmt.Errorf("unable to read dependency mappings file\n%w, err)`

dependency_mapping.go
```diff
@@ -0,0 +1,54 @@
...
+	_, err := toml.DecodeFile(path, &mappingsFile)
+	if err != nil {
+		return MappingsFile{}, err
```
nebhale, line 47: More descriptive error: `fmt.Errorf("unable to decode dependency mappings file%s\n%w, path, err)`

dependency_cache.go
```diff
@@ -48,16 +48,31 @@ type DependencyCache struct {
...
+	}
+	var mappings []DependencyMapping
+	for _, bpm := range mappingsFile.BuildpackMappings {
+		if bpm.BuildpackID == context.Buildpack.Info.ID {
+			mappings = bpm.Mappings
```
nebhale, line 67: Would it be clearer to create the struct first and do a direct assignment?

```go
dc := DependencyCache {
...
}

...

if bpm.BuildpackID == context.Buildpack.Info.ID {
  dc.Mappings = bpm.Mappings
}

return dc, nil
```

nebhale, line 67: There should also be a log message (not sure of how prominent we want it to be) that mentions that mappings are going to be used for this buildpack.

dependency_mapping.go
ekcasey: I had that originally but I thought somebody could realistically want to use this library to write the file in the correct format

ekcasey: happy to change it back if you think that is unrealistic

nebhale: Yeah, I wouldn't optimize for that right now.  It's a good thought, but I'm pretty realistic that we're the only ones that are going to use this.

dependency_cache.go
ekcasey: Omitting this for now because the Logger is not injected into the constructor

PR #16, opened 2020-10-19
netrc

nebhale:
This change adds support for reading a [netrc][1] file for basic auth credentials when downloading a buildpack's dependency.  It also makes some modifications to the download layer API to simplify it using a variadic argument.

[1]: https://www.gnu.org/software/inetutils/manual/html_node/The-_002enetrc-file.html


nebhale:
@ekcasey Please cut a release of `libpak` once this merges in.  Thanks!

ekcasey (approved):
Looks great! It's probably worth documenting this feature of the packager somewhere in the future.

PR #22, opened 2020-11-02
Introduces a new interface for resolving bindings.

ekcasey:
The previous BindingsResolver did not provide an interface for resolving multiple bindings of a given type. Some paketo use-cases such as dependency-mappings and ca-certficates should be able to support an arbitrary number of bindigns. This changes intoduces a new bindings package that exports both Resolve and ResolveOne. In addition it provides functionality that allows the consumer to filter on provider in addition to type.

Signed-off-by: Emily Casey <ecasey@vmware.com>


bindings/resolve.go
```diff
@@ -0,0 +1,89 @@
...
+
+// OfType returns a Predicate that returns true if a given binding has Type that matches t. The comparison is
+// case-insensitive.
+func OfType(t string) Predicate {
+	return func(bind libcnb.Binding) bool {
+		return strings.ToLower(bind.Type) == strings.ToLower(t)
+	}
+}
+
+// OfType returns a Predicate that returns true if a given binding has Provider that matches p. The comparison is
```
nebhale, line 37:
```suggestion
// OfProvider returns a Predicate that returns true if a given binding has Provider that matches p. The comparison is
```

nebhale (approved):

PR #31, opened 2020-11-13
Helper Helpers

nebhale:
This change adds some more helper functions to sherpa that are particularly useful in the environment that exec.d's run in.


bindings/resolve.go
```diff
@@ -26,6 +26,36 @@ import (
 // Predicate should return true if it matches a given binding.
 type Predicate func(bind libcnb.Binding) bool
 
+// And combines multiple predicates logical and-ing their results together.
+func And(predicates ...Predicate) Predicate {
```
ekcasey, line 30: Do we need this one? `And` is the default behavior when multiple Predicates are provided.

sherpa/env_var_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+			it("appends multiple", func() {
+				Expect(sherpa.AppendToEnvVar("TEST_KEY", "|", "test-value-2", "test-value-3")).
+					To(Equal("test-value-2|test-value-3"))
+			})
+		})
+
+		context("With Existing", func() {
+			it.Before(func() {
+				Expect(os.Setenv("TEST_KEY", "test-value-1")).To(Succeed())
```
ekcasey, line 51: The tests are so fast already that is not a big win, but worth noting, that `spec` is capable of running tests and parallel. We can't take advantage of that now because we use the same global state in multiple tests. If we faked the env in our tests or used randomized key name (would need to be done in several additional place) we could start running the tests in parallel.

sherpa/env_var.go
```diff
@@ -0,0 +1,52 @@
...
+	if s, ok := os.LookupEnv(name); ok {
+		e = append(e, s)
+	}
+	e = append(e, values...)
+	return strings.Join(e, delimiter)
+}
+
+// GetEnvRequired returns the value of an environment varible if it exists, otherwire returns an error with a
+// predictable message.
+func GetEnvRequired(name string) (string, error) {
```
ekcasey, line 38: I see the value of the other two, but why would I as a consumer choose to use `GetEnvRequired` instead of `os.LookupEnv`? It doesn't seem like it is saving any work other than a couple words in the error message.

sherpa/env_var.go
```diff
@@ -0,0 +1,52 @@
...
+func AppendToEnvVar(name string, delimiter string, values ...string) string {
+	var e []string
+	if s, ok := os.LookupEnv(name); ok {
+		e = append(e, s)
+	}
+	e = append(e, values...)
+	return strings.Join(e, delimiter)
+}
+
+// GetEnvRequired returns the value of an environment varible if it exists, otherwire returns an error with a
```
ekcasey, line 36:
```suggestion
// GetEnvRequired returns the value of an environment variable if it exists, otherwise returns an error with a
```

bindings/resolve.go
nebhale: Nope, fair enough.

sherpa/env_var.go
nebhale: I think it is worth it when the difference is:

```go
agentPath, ok := os.LookupEnv("BPI_GOOGLE_STACKDRIVER_DEBUGGER_JAVA_AGENT_PATH")
if !ok {
	return nil, fmt.Errorf("$BPI_GOOGLE_STACKDRIVER_DEBUGGER_JAVA_AGENT_PATH must be set")
}
```
and
```go
agentPath, err := sherpa.GetEnvRequired("BPI_GOOGLE_STACKDRIVER_DEBUGGER_JAVA_AGENT_PATH")
if err != nil {
	return nil, err
}
```

ekcasey (approved):

nebhale:
@ekcasey I’m going to leave this in draft until we’ve got a proto-typical implementation of your preferred binding-related buildpack design.  Just in case we need to add or remove anything else.

ekcasey (approved):

dmikusa-pivotal (approved):

PR #73, opened 2021-08-16
Adds convenience method to generate a file listing & return the hash of the listing

### LINKED ISSUE #25: FileLists and MavenFileLists Metadata is too big
A user has reported that the file lists and maven file list metadata used for some layer caches is bigger than it needs to be.  The suggestion from this user was that we hash the file list once created and store that instead.  That will give us the same behavior in a much smaller footprint.

dmikusa-pivotal:
## Summary

The file listing mechanism is typically used to detect if changes have occurred under a directory. Rather than have the full details of the file listing, it's sufficient to compare hashes of the file listings. The hash also results in a smaller value, if you need to store that in metadata.

## Use Cases

Resolves #25 

## Checklist

### DIFF
--- a/sherpa/file_listing.go
+++ b/sherpa/file_listing.go
@@ -46,6 +46,21 @@ type result struct {
 	value FileEntry
 }
 
+// NewFileListingHash generates a sha256 hash from the listing of all entries under the roots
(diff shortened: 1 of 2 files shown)


pivotal-david-osullivan (approved):
LGTM!

PR #115, opened 2021-12-14
Remove unnecessary coversion from XML > JSON for CycloneDx + syft

samj1912:
Signed-off-by: Sambhav Kothari <skothari44@bloomberg.net>

Since we have updated to syft 0.32 - syft natively supports cyclonedxjson format. We should use it directly.

https://github.com/paketo-buildpacks/java/blob/25952ed3b1bd77d5ef8374f1f3ba1163249dea24/buildpack.toml#L46

https://github.com/paketo-buildpacks/syft/releases/tag/v1.2.0

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/paketo-buildpacks/libpak
 go 1.15
 
 require (
-	github.com/CycloneDX/cyclonedx-go v0.4.0
 	github.com/Masterminds/semver/v3 v3.1.1
 	github.com/buildpacks/libcnb v1.25.2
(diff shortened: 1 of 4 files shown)


dmikusa-pivotal:
Thanks so much! I had this on my TODO list and you beat me to it.

One thing, I had expected this to remove `github.com/CycloneDX/cyclonedx-go v0.4.0` from go.mod as that's the only place we were using that library. I don't see that in the PR though? Can you try a `go mod tidy` or something and see if that clears it out?

dmikusa-pivotal (approved):

PR #128, opened 2022-03-02
Add a helper to extract an archive based on mime type

garethjevans:
## Summary
A helper function to extract an archive based on file type

## Use Cases
Determines the mime type of the file, and uses the correct method to extract the archive, this will help avoid lots of `if else if ...` logic in various method.

## Checklist

### DIFF
--- a/crush/crush.go
+++ b/crush/crush.go
@@ -23,11 +23,13 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/xi2/xz"
+	"github.com/h2non/filetype"
 )
 
 // CreateTar writes a TAR to the destination io.Writer containing the directories and files in the source folder.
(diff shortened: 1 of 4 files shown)


garethjevans:
@dmikusa-pivotal fyi

crush/crush.go
```diff
@@ -101,6 +103,38 @@ func CreateTarGz(destination io.Writer, source string) error {
...
+		return fmt.Errorf("unknown file type")
+	}
+
+	switch kind.MIME.Value {
+	case "application/x-tar":
+		return ExtractTar(source, destination, stripComponents)
+	case "application/x-bzip2":
+		return ExtractTarBz2(source, destination, stripComponents)
+	case "application/gzip":
+		return ExtractTarGz(source, destination, stripComponents)
```
dmikusa-pivotal, line 128: This assumes that you have a `.tar.gz` or `.tar.bz2` file. You can in fact just have a `.gz` file, like if you gzip a go binary executable.

I think this needs to be handled differently. My thought would be to have an internal `extractArchive(source io.Reader, destination string, stripComponents int)` method. I think you could then do something like this:

```
case "application/x-bzip2":
    return extractArchive(bzip2.NewReader(source), destination, stripComponents)
...
```

Basically, pass in a reader that unzip's it and look at what's in that file. I think you can use [MatchReader](https://github.com/h2non/filetype/blob/master/match.go#L55) instead of Match and it should be able to read from the reader and match against that.

crush/crush.go
```diff
@@ -101,6 +103,38 @@ func CreateTarGz(destination io.Writer, source string) error {
...
+// components can be stripped from each path.
+func ExtractArchive(source *os.File, destination string, stripComponents int) error {
+	b, err := ioutil.ReadFile(source.Name())
+	if err != nil {
+		return err
+	}
+	kind, err := filetype.Match(b)
+	if err != nil {
+		return err
+	}
```
dmikusa-pivotal, line 116: As I read this, it's loading the whole file into memory and passing that to Match. I don't think we'd want to do that because you have some archives which can be up to 500M. It's probably better to use one of the other matchers, like [MatchReader](https://github.com/h2non/filetype/blob/master/match.go#L55-L64) which seems to only read 8K.

dmikusa-pivotal (changes requested):

dmikusa-pivotal:
I was playing around with this and wanted to make some changes. It seemed a little easier to branch off what you have here and make the changes directly so I submitted https://github.com/paketo-buildpacks/libpak/pull/132. Can you take a look at that and tell me if that works for you?

I think I've solved the issue of loading the full archive into memory by using a buffer/TeeReader/MultiReader. In my tests, it only loads at most the first 8192 bytes into memory. There are some other changes in there as well though, so let me know what you think or if I've missed anything for your use case. Thanks!

dmikusa-pivotal:
I was playing around with this and wanted to make some changes. It seemed a little easier to branch off what you have here and make the changes directly so I submitted https://github.com/paketo-buildpacks/libpak/pull/132. Can you take a look at that and tell me if that works for you?

I think I've solved the issue of loading the full archive into memory by using a buffer/TeeReader/MultiReader. In my tests, it only loads at most the first 8192 bytes into memory. There are some other changes in there as well though, so let me know what you think or if I've missed anything for your use case. Thanks!

dmikusa-pivotal:
Closing in favor of #132

PR #132, opened 2022-03-16
Consolidate extract file methods

dmikusa-pivotal:
## Summary

Consolidate the different extract file methods into one `Extract` method that uses the mime-types for determining the file type.

This PR also deprecates the old extract individual file type methods with the goal of having one method `Extract` so a buildpack author doesn't need to care about individual archive file types.

## Use Cases

In some cases, a buildpack only needs a specific file format, so it's not too much of a pain, but if a buildpack needs to support extracting files from multiple different archive formats (say zip, tar.gz, tar.bz2, etc..) what ends up happening is that you get a large `if..else if...else if..` block (or a switch statement) going in the code which is ugly. This also tends to use file extensions for the logic and using mime types is a better indicator of the actual file type than an extension.

## Checklist


garethjevans:
@dmikusa-pivotal those changes work for me, I notice for the zip extraction we are still reading the whole file into the buffer, but I couldn't find a way around that either.

PR #141, opened 2022-04-25
Rewrite LayerContributor

### LINKED ISSUE #138: Layer contributors should handle when an uncached layer becomes cached
## Current Behavior
**Given** A layer from a previous build that was `cache=false` and `launch=true`
**Given** A buildpack wants to contribute a layer that is identical in every way *except* that `cache=false`
**When** The buildpack tries to use a `LayerContributor` to provide this layer
**Then** The resulting layer is empty, resulting in a failed build

```
===> EXPORTING
Reusing layer 'paketo-buildpacks/ca-certificates:helper'
Adding layer 'paketo-buildpacks/bellsoft-liberica:helper'
Adding layer 'paketo-buildpacks/bellsoft-liberica:java-security-properties'
ERROR: failed to export: layer 'paketo-buildpacks/bellsoft-liberica:jre' is cache=true but has no contents
```

## Expected Behavior
If the expected layer is cached, the LayerContributor should check for layer contents and contribute the layer if none exist.

dmikusa-pivotal:
## Summary

- Check if the layer has been restored properly. If the layer is cache or build and the layer TOML exists, then the layer directory should also exist. If it does not, then the layer was not restored and we need to reload the laery. This adds checks for this condition and will print a different message to indicate that the layer is being reloaded, instead of contributed fresh.

- Add sherpa utilites for checking if a file, directory or symlink exist. These were needed for the refactoring above and are also common across other buildpacks. A set of tests is also provided.

- Fixes test "adds expected Syft SBOM file" which was previously passing because a previous test run creates the SBOM file this test requires. Refactoring of the test class exposed this error. The test has been corrected.

## Use Cases

Resolves #138 

## Checklist


layer.go
```diff
@@ -105,6 +103,60 @@ func (l *LayerContributor) Contribute(layer libcnb.Layer, f LayerFunc) (libcnb.L
...
+	l.Logger.Debugf("Actual metadata: %+v", layer.Metadata)
+
+	return expected, reflect.DeepEqual(expected, layer.Metadata), nil
+}
+
+func (l *LayerContributor) checkIfLayerRestored(layer libcnb.Layer) (bool, error) {
+	layerTOML := fmt.Sprintf("%s.toml", layer.Path)
+	tomlExists, err := sherpa.FileExists(layerTOML)
+	if err != nil {
+		return false, fmt.Errorf("unable to check if layer TOML tomlExists %s\n%w", layerTOML, err)
```
pivotal-david-osullivan, line 127:
```suggestion
		return false, fmt.Errorf("unable to check if layer toml exists %s\n%w", layerTOML, err)
```

pivotal-david-osullivan (approved):

PR #179, opened 2022-10-06
Display deprecated dependencies during build

modulo11:
## Summary

As briefly discussed in [Slack](https://paketobuildpacks.slack.com/archives/C0124SD3GTG/p1663854580399979), this changes displays deprecated (or soon to be) dependencies during build. This is inline with the implementation done in packit and e.g. used in the `node-engine` buildpacks.

It currently looks like this:

```
Paketo Buildpack for SapMachine {{.version}}
  https://github.com/paketo-buildpacks/sap-machine
  Build Configuration:
  ...
  Deprecation Notice:
    Version 11.0.16 of SapMachine JDK is deprecated.
    Migrate your application to a supported version of SapMachine JDK.
  Deprecation Notice:
    Version 11.0.16 of SapMachine JRE is deprecated.
    Migrate your application to a supported version of SapMachine JRE.
  SapMachine JDK 11.0.16: Contributing to layer
```

This change should be non breaking and require minimal changes e.g. in `libjvm`. 

Alternatively the functionality could also be provided separately to the `Resolve` method. I'd leave it to the reviewer to decide this.

## Use Cases

- Display deprecated (soon to be deprecated) dependencies during build time

## Checklist


buildpack.go
```diff
@@ -479,7 +507,13 @@ func (d *DependencyResolver) Resolve(id string, version string) (BuildpackDepend
 		return a.GreaterThan(b)
 	})
 
-	return candidates[0], nil
+	candidate := candidates[0]
+
+	if (candidate.DeprecationDate != time.Time{}) {
+		d.printDependencyDeprecation(candidate)
+	}
```
dmikusa, line 514: I would like to see a test for this in `buildpack_test.go`. The section that tests the dependency resolver could do this. There are test cases now that do not set the deprecation date, so you could use one of those to validate that nothing is logged.

Then add a test case (or modify a different existing test case) to include the deprecation date & validate that output is logged.

You can pass the `bard.Logger` to the dependency resolver, your change adds that. If you pass in a `&bytes.Buffer{}` and then fetch the logged output (or lack of output) from that.

dmikusa (changes requested):
Overall it looks good to me. One small suggested change, to just add test coverage for the Resolve method changes. See my additional comment.

In regards to integrating with JVM based buildpacks, what steps do you see necessary? I think we'd need to bump the libpak version, and modify buildpack.toml to include this dependency info. Is there anything I'm missing? You'd mentioned libjvm, and it could be cause it's late at the moment, but I'm not sure what required changes would be needed there.

Thanks!

modulo11:
This would be my proposed plan:

1. Merge PR and cut a new release of `libpak`
2. Add `deprecation_date` to `buildpack.toml` (does this need changes to the version bumping mechanism to ignore `deprecation_date`?)
3. Update `libpak` in [libjvm](https://github.com/paketo-buildpacks/libjvm) (and other places where it's used) and apply this patch:
```diff
index 43a4400..ce83234 100644
--- a/build.go
+++ b/build.go
@@ -111,7 +112,7 @@ func (b Build) Build(context libcnb.BuildContext) (libcnb.BuildResult, error) {
                return libcnb.BuildResult{}, fmt.Errorf("unable to determine jvm version\n%w", err)
        }
 
-       dr, err := libpak.NewDependencyResolver(context)
+       dr, err := libpak.NewDependencyResolver(context, b.Logger)
        if err != nil {
                return libcnb.BuildResult{}, fmt.Errorf("unable to create dependency resolver\n%w", err)
        }
```
4. Release and bump `libjvm`

Because `deprecation_date` is only optionally displayed, step 2 could be implemented at any time.

buildpack.go
```diff
@@ -240,6 +256,16 @@ func NewBuildpackMetadata(metadata map[string]interface{}) (BuildpackMetadata, e
 				d.PURL = v
 			}
 
+			if v, ok := v["deprecation_date"].(string); ok {
+				deprecationDate, err := time.Parse(time.RFC3339, v)
+
+				if err != nil {
+					return BuildpackMetadata{}, err
```
pivotal-david-osullivan, line 263:
```suggestion
					return BuildpackMetadata{}, fmt.Errorf("unable to parse deprecation date\n%w", err)
```
Small change to wrap error

buildpack.go
```diff
@@ -412,16 +438,18 @@ type DependencyResolver struct {
 
 	// StackID is the stack id of the build.
 	StackID string
+
+	logger bard.Logger
 }
 
 // NewDependencyResolver creates a new instance from the buildpack metadata and stack id.
-func NewDependencyResolver(context libcnb.BuildContext) (DependencyResolver, error) {
+func NewDependencyResolver(context libcnb.BuildContext, logger bard.Logger) (DependencyResolver, error) {
```
pivotal-david-osullivan, line 446: cc: @dmikusa 
Wouldn't this break all calls to NewDependencyResolver?

Could we instead leave the logger added to the DependencyResolver struct (changing to a public property with logger -> Logger) and have users initialize it when necessary via:

```
dr, err := libpak.NewDependencyResolver(context)
...
dr.Logger = <b.Logger>
```

modulo11: Yes, it would need some adjustments. I've seen that in the code multiple times, but I have to admit that I don't like it very much. The constructor should bring an objects into a usable state. 

Alternatively we could extract the functionality and do something like that (on the caller side):

```go
checker := libpak.NewDependencyChecker(&b.Logger)

depJDK, err := dr.Resolve("jdk", v)
if (jdkRequired && !nativeImage) && err != nil {
   return libcnb.BuildResult{}, fmt.Errorf("unable to find dependency\n%w", err)
}

checker.Check(depJDK)
```

dmikusa: I agree it's best if the constructor makes a usable object, but we can't break the API. There are also several place where the Logger is set in this manner. To keep the API stable, I'd also be fine doing that here. It's definitely on my todo list so this is changed in v2.

Other options:

1. Make a totally new `NewDependencyResolverWithLogger`.
2. Use the [option pattern](https://github.com/paketo-community/cargo/blob/9a30f3c5983f7e3b03bde63f1cae4b5d88c376cc/cargo/cargo.go#L36). `NewDependencyResolver(context libcnb.BuildContext, options ...Option)`.  This way the options are not required so the API is consistent, but one can pass through things like the logger. It doesn't require one to pass through anything that optional though, which may not really solve the problem here. If a Logger is required, then a user could still forget to include it and cause problems.

modulo11: That makes sense and as it's already on your list for `v2`, I'm fine with using the easiest route.

dmikusa: And FYI, I think we'll probably shoot for a v2 after libcnb releases it's v2. That way libpak v2 matches with libcnb v2.

modulo11:
Is there anything left to do? The pending change request should be addressed.

dmikusa:
>(does this need changes to the version bumping mechanism to ignore deprecation_date?)

It should not. The tool only touches fields that it specifically modifies. Everything else should be echoed through.

dmikusa:
I tested this. The plan described worked for me and I was able to get the deprecation notices.

The only issue I came upon is that this treats everything as UTC. We are specifying deprecation *dates* but Go treats them as date & time. So I put in tomorrow, thinking it would give me the "deprecated soon" warning and it did not. What happened is that I'm 4 hours behind UTC and four hours from now (i.e. when I'm testing) is actually tomorrow so I got the deprecated message, not the deprecated soon message.

I think in a perfect world, the times entered in the buildpack.toml would be in UTC. Then when doing the comparison the buildpack would look up the local timezone and make the adjustment when comparing the times. I don't know the level of effort on doing that or how much code it would add, like if it would require another dependency.

My vote would be that if we can do it easily, without adding another dependency or significant amounts of code, then we should. Otherwise, I think we should just make a note about times being in UTC and call it a day.

Thoughts @pivotal-david-osullivan ?

modulo11:
> 



> I tested this. The plan described worked for me and I was able to get the deprecation notices.
> 
> The only issue I came upon is that this treats everything as UTC. We are specifying deprecation _dates_ but Go treats them as date & time. So I put in tomorrow, thinking it would give me the "deprecated soon" warning and it did not. What happened is that I'm 4 hours behind UTC and four hours from now (i.e. when I'm testing) is actually tomorrow so I got the deprecated message, not the deprecated soon message.
> 
> I think in a perfect world, the times entered in the buildpack.toml would be in UTC. Then when doing the comparison the buildpack would look up the local timezone and make the adjustment when comparing the times. I don't know the level of effort on doing that or how much code it would add, like if it would require another dependency.
> 
> My vote would be that if we can do it easily, without adding another dependency or significant amounts of code, then we should. Otherwise, I think we should just make a note about times being in UTC and call it a day.
> 
> Thoughts @pivotal-david-osullivan ?

I'll check that out. As the implementation is largely taken from packit, I'll check that as well afterwards.

dmikusa (approved):

dmikusa:
Thanks for the change to UTC. I've approved it, but Github is telling me the branch is outdated. Usually, it allows me to update the branch, but not this time. Can you update the branch? Then I can merge it. Thanks

PR #237, opened 2023-04-28
Add Predicate `WithName` to filter bindings

c0d1ngm0nk3y:
see https://github.com/buildpacks/libcnb/pull/228

## Summary

Being able to get a binding with a specific name could be useful in specific situation.

## Use Cases

If the bindings are read from `VCAP_SERVICES` and the binding is a "user-provided-service", then the type would always be `user-provided`. So resolving it by type makes no sense, but resolving it by name should be fine since the user can choose the name when creating a "user-provided-service".

## Checklist

### DIFF
--- a/bindings/resolve.go
+++ b/bindings/resolve.go
@@ -30,7 +30,7 @@ type Predicate func(bind libcnb.Binding) bool
 // case-insensitive.
 func OfType(t string) Predicate {
 	return func(bind libcnb.Binding) bool {
-		return strings.ToLower(bind.Type) == strings.ToLower(t)
+		return strings.EqualFold(bind.Type, t)
 	}
 }
 
@@ -38,7 +38,15 @@ func OfType(t string) Predicate {
 // case-insensitive.
 func OfProvider(p string) Predicate {
 	return func(bind libcnb.Binding) bool {
-		return strings.ToLower(bind.Provider) == strings.ToLower(p)
+		return strings.EqualFold(bind.Provider, p)
+	}
+}
+
+// WithName returns a Predicate that returns true if a given binding has Name that matches n. The comparison is
+// case-insensitive.
+func WithName(n string) Predicate {
+	return func(bind libcnb.Binding) bool {
+		return strings.EqualFold(bind.Name, n)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


c0d1ngm0nk3y:
@paketo-buildpacks/java-maintainers What do you think about this?

bindings/resolve.go
```diff
@@ -30,25 +30,32 @@ type Predicate func(bind libcnb.Binding) bool
 // case-insensitive.
 func OfType(t string) Predicate {
 	return func(bind libcnb.Binding) bool {
-		return strings.ToLower(bind.Type) == strings.ToLower(t)
+		return strings.EqualFold(bind.Type, t)
```
dmikusa, line 33: I'm curious about this change. Are you seeing specific issues with `ToLower`? The behavior isn't the same, so wanting to understand more.

dmikusa (commented):
Just one question.

dmikusa:
Oh, also not sure if you saw my comment on the related libcnb PR: https://github.com/buildpacks/libcnb/pull/228#issuecomment-1537176590

bindings/resolve.go
c0d1ngm0nk3y: It was just some issue raised by a linter in the `IDE`.

> strings.EqualFold() Function in Golang reports whether s and t, interpreted as UTF-8 strings, are equal under Unicode case-folding, which is a more general form of case-insensitivity. 

For non UTF-8, it should be the same.

c0d1ngm0nk3y:
> Oh, also not sure if you saw my comment on the related libcnb PR: [buildpacks/libcnb#228 (comment)](https://github.com/buildpacks/libcnb/pull/228#issuecomment-1537176590)

But I think this would still be valid in case of user user provided services since `provider` **AND** `type` would be `user-provided` and the user can only influence the name.

dmikusa:
> > Oh, also not sure if you saw my comment on the related libcnb PR: [buildpacks/libcnb#228 (comment)](https://github.com/buildpacks/libcnb/pull/228#issuecomment-1537176590)
> 
> But I think this would still be valid in case of user user provided services since `provider` **AND** `type` would be `user-provided` and the user can only influence the name.

Ok, perfect. I just wanted to draw your attention to that other comment. I agree, for this case it is the same.

dmikusa (approved):

PR #240, opened 2023-05-12
Do not check DeprecationDate when comparing dependencies

c0d1ngm0nk3y:
## Summary

This might be an alternative to #233 in addressing short comings when comes to comparing the metadata of a dependency.

* `DeprecationDate` does not work with `reflect.deepEqual` for 2 reasons
  * After `Marshal` and `Unmarshal`, the `time.Time` cannot be compared anymore, since it will loose some precision because of the missing `UnmarshalTOML`
  * Even **WITH** the fixed unmarshaling, the compare will work, but the `reflect.deepEqual` will still fail because it compares the exact internal structure.
* If the `CPEs` array is empty, those a arrays are not considered to be `reflect.deepEqual`   

## Use Cases

As discussed in #233, comparing the metadata of a dependency is probably not needed at all if `sha256` is the same. But at least the comparison should be successful in those cases since the dependency caching currently depends on it.

## Checklist


buildpack.go
```diff
@@ -95,6 +96,23 @@ type BuildpackDependency struct {
...
+	b2.DeprecationDate = time.Time{}
+
+	if len(b1.CPEs) == 0 {
+		b1.CPEs = nil
+	}
+	if len(b2.CPEs) == 0 {
+		b2.CPEs = nil
+	}
+
+	return reflect.DeepEqual(b1, b2)
```
dmikusa, line 113:
```suggestion
	b1.DeprecationDate = b1.DeprecationDate.Truncate(time.Second).In(time.UTC)
	b2.DeprecationDate = b2.DeprecationDate.Truncate(time.Second).In(time.UTC)

	if len(b1.CPEs) == 0 {
		b1.CPEs = nil
	}
	if len(b2.CPEs) == 0 {
		b2.CPEs = nil
	}

	return reflect.DeepEqual(b1, b2)
```

I think if we truncate to the second & convert to UTC we can normalize the depreciation dates but still get a reasonable comparison. We should need to really even compare to the second, since it's just a *date*, but not sure if we want to go that far.

Thoughts?

c0d1ngm0nk3y: Like discussed in https://github.com/paketo-buildpacks/libpak/pull/233, I do not really see the point in comparing the metadata in the first place. And when it comes to `deprecationDate`, it never worked anyway. But that fixes the issue as well (I still think overly complicated, but it fixes it) and the cache misses are really annoying when downloading the same `JDK` over and over again.

modulo11:
This looks like a cost-efficient and compatible alternative until #233 can be continued on a v2 branch. @dmikusa is there anything holding us back from merging it?

dmikusa (approved):

PR #267, opened 2023-07-07
Introduce struct with metadata to identify a dependency

c0d1ngm0nk3y:
related to https://github.com/paketo-buildpacks/libpak/pull/233#issuecomment-1528865966

## Summary

This introduces a Metadata struct with all relevant properties to identify a dependency and still be descriptive.

## Use Cases

This allow users (e.g. `libjvm`) to use this metadata as layer metadata and not the whole dependency. That would
a) Simplify comparison if there is a cache hit (special logic for `deprecation_data`)
b) Ensure that no cache miss happens because some metadata changed (e.g. `url`)  

## Checklist

### DIFF
--- a/buildpack.go
+++ b/buildpack.go
@@ -96,8 +96,8 @@ type BuildpackDependency struct {
 	DeprecationDate time.Time `toml:"deprecation_date"`
 }
 
-// BuildpackDependencyMetadata describes the dependency
-type BuildpackDependencyMetadata struct {
+// DependencyLayerContributorMetadata returns the subset of data from BuildpackDependency that is use as expected metadata for the DependencyLayerContributor.
+type DependencyLayerContributorMetadata struct {
 	// ID is the dependency ID.
 	ID string `toml:"id"`
 
--- a/buildpack.go
+++ b/buildpack.go
@@ -113,7 +113,7 @@ type DependencyLayerContributorMetadata struct {
 
 // GetMetadata return the relevant metadata of this dependency
 func (b BuildpackDependency) GetMetadata() DependencyLayerContributorMetadata {
-	return BuildpackDependencyMetadata{
+	return DependencyLayerContributorMetadata{
 		ID:      b.ID,
 		Name:    b.Name,
 		Version: b.Version,
--- a/buildpack.go
+++ b/buildpack.go
@@ -112,7 +112,7 @@ type DependencyLayerContributorMetadata struct {
 }
 
 // GetMetadata return the relevant metadata of this dependency
-func (b BuildpackDependency) GetMetadata() BuildpackDependencyMetadata {
+func (b BuildpackDependency) GetMetadata() DependencyLayerContributorMetadata {
 	return BuildpackDependencyMetadata{
 		ID:      b.ID,
 		Name:    b.Name,
--- a/buildpack.go
+++ b/buildpack.go
@@ -96,6 +96,31 @@ type BuildpackDependency struct {
 	DeprecationDate time.Time `toml:"deprecation_date"`
 }
 
+// BuildpackDependencyMetadata describes the dependency
+type BuildpackDependencyMetadata struct {
+	// ID is the dependency ID.
+	ID string `toml:"id"`
+
+	// Name is the dependency name.
+	Name string `toml:"name"`
+
+	// Version is the dependency version.
+	Version string `toml:"version"`
+
+	// SHA256 is the hash of the dependency.
+	SHA256 string `toml:"sha256"`
+}
+
+// GetMetadata return the relevant metadata of this dependency
+func (b BuildpackDependency) GetMetadata() BuildpackDependencyMetadata {
+	return BuildpackDependencyMetadata{
+		ID:      b.ID,
+		Name:    b.Name,
+		Version: b.Version,
+		SHA256:  b.SHA256,
+	}
+}
+
 // Equals compares the 2 structs if they are equal. This is very simiar to reflect.DeepEqual
 // except that properties that will not work (e.g. DeprecationDate) are ignored.
 func (b1 BuildpackDependency) Equals(b2 BuildpackDependency) bool {
--- a/layer.go
+++ b/layer.go
@@ -237,7 +237,7 @@ func NewDependencyLayer(dependency BuildpackDependency, cache DependencyCache, t
 func NewDependencyLayerContributor(dependency BuildpackDependency, cache DependencyCache, types libcnb.LayerTypes) DependencyLayerContributor {
 	return DependencyLayerContributor{
 		Dependency:       dependency,
-		ExpectedMetadata: dependency,
+		ExpectedMetadata: dependency.GetMetadata(),
 		DependencyCache:  cache,
 		ExpectedTypes:    types,
 	}


dmikusa:
I'm not 100% sure I follow here. It looks like this PR is adding a method that will return a subset of the dependency's metadata for purposes of putting that into the layer contributor's metadata, which gets used for comparison/caching. If I'm understanding that correctly, it sounds good to me. I do have a couple of naming suggestions I'll add in different comments.

buildpack.go
```diff
@@ -96,6 +96,31 @@ type BuildpackDependency struct {
 	DeprecationDate time.Time `toml:"deprecation_date"`
 }
 
+// BuildpackDependencyMetadata describes the dependency
+type BuildpackDependencyMetadata struct {
```
dmikusa, line 100:
```suggestion
// LayerContributorDependencyMetadata returns the subset of data from BuildpackDependency that is use as expected metadata for the DependencyLayerContributor.
type DependencyLayerContributorMetadata struct {
```

c0d1ngm0nk3y:
> I'm not 100% sure I follow here.

Fair enough. I had to leave, so my description felt a bit short. There are basically 3 steps:

1) Providing this structure to be able to filter the relevant metadata. First I wanted to go only for `sha256`, but then the metadata is not very descriptive.
2) Consume it in `libjvm` to set the metadata for a layer (e.g. [here](https://github.com/paketo-buildpacks/libjvm/blob/75bc1c9b0174b3b440b94c1a0e416b59c5d0d9c1/jdk.go#L39))
3) This allows to simplify the comparison of the metadata in `libpak` and get rid of [this](https://github.com/paketo-buildpacks/libpak/blob/ce48dbc5ae2cbe5e26c1022fdd32690200ad8829/layer.go#L129) very specific logic. There should not be need for this anymore.

WDYT?

dmikusa:
That sounds reasonable to me. A couple of clarifying questions...

>First I wanted to go only for sha256, but then the metadata is not very descriptive.

Is the plan to still only compare on the sha256 hash? but to have this additional metadata for logging/output purposes?

>This allows to simplify the comparison of the metadata in libpak and get rid of [this](https://github.com/paketo-buildpacks/libpak/blob/ce48dbc5ae2cbe5e26c1022fdd32690200ad8829/layer.go#L129) very specific logic. There should not be need for this anymore.

Would that allow us to stop using `reflect.DeepEqual` for comparison? 

This has been a source of confusion. When it returns false, it's never clear why and can sometimes. I'm guessing yes, since, going forward, we're just comparing the sha256 hashes.

c0d1ngm0nk3y:
I have added the suggestions and once this is merged, released and consumed, `libjvm` can consume it.

layer.go
```diff
@@ -237,7 +237,7 @@ func NewDependencyLayer(dependency BuildpackDependency, cache DependencyCache, t
 func NewDependencyLayerContributor(dependency BuildpackDependency, cache DependencyCache, types libcnb.LayerTypes) DependencyLayerContributor {
 	return DependencyLayerContributor{
 		Dependency:       dependency,
-		ExpectedMetadata: dependency,
+		ExpectedMetadata: dependency.GetMetadata(),
```
dmikusa, line 240: ExpectedMetadata on the DependencyLayerContributor [is an interface{}](https://github.com/paketo-buildpacks/libpak/pull/267/files#diff-c624765f4df697256be192f9789ec38a15b70661f77956b29d03c81856fa8414R206) so I'm not sure I follow how adding a type here is going to help. We're going to immediately throw away that type.

Should we be changing ExpectedMetadata to have a type? but that could be considered a breaking change because ExpectedMetadata is a public field, so I'm not sure we can change the type.

Am I missing something here? Are you just going to cast this in libjvm?

c0d1ngm0nk3y:
> Is the plan to still only compare on the sha256 hash? but to have this additional metadata for logging/output purposes?

I think, I would keep using `reflect.DeepEqual` since the metadata could contain anything. True, `ID`, `Name` or `Version` could change with `SHA256` being stable, but that sounds rather academic to me. Maybe in some odd edge cases, but probably we can live with that.

> > This allows to simplify the comparison of the metadata in libpak and get rid of [this](https://github.com/paketo-buildpacks/libpak/blob/ce48dbc5ae2cbe5e26c1022fdd32690200ad8829/layer.go#L129) very specific logic. There should not be need for this anymore.
> 
> Would that allow us to stop using `reflect.DeepEqual` for comparison?


> This has been a source of confusion. When it returns false, it's never clear why and can sometimes. I'm guessing yes, since, going forward, we're just comparing the sha256 hashes.

In this specific case, yes. But `ExpectedMetadata` is `interface{}` and could be anything, so I think `reflect.DeepEqual` is the right choice. But as long as `ExpectedMetadata` is only a simple and small `struct`, we should be fine. And this `struct` won't change if something got added to the `Dependency` over time.

c0d1ngm0nk3y:
The usage would look like https://github.com/paketo-buildpacks/libjvm/pull/312.

After that, we don't need [this coding](https://github.com/paketo-buildpacks/libpak/blob/37ca34bdcfbe94b971f7c12bdbd616619dba73bc/layer.go#L129) (`Equals`) anymore.

Did I miss something?

layer.go
c0d1ngm0nk3y: > Are you just going to cast this in libjvm?

No, but assigning it to `interface{}` should be fine, shouldn't it? The type information is not really important since we would still use `reflect.DeepEqual`. But there are only 4 strings in this struct to compare. See https://github.com/paketo-buildpacks/libjvm/pull/312 on how the change in `libjvm` would look like.

dmikusa: I'm just struggling with the risk/benefit on this one. I get that we're only comparing a subset of metadata, and so that should make the comparison work better, but on the other hand we've already fixed the comparison in a different way. It's not exactly an elegant solution, but it works. I'm just hesitant to change the code again, which always introduces some risk as well as maintenance burden pushing releases through everywhere.

Especially since I can say with 99.99% certainty that I would not approve a PR against v1 that does change the metadata structure at this point. Anything like that should go into v2.

I wouldn't be opposed to doing something like this in v2. I think we can definitely do better in this regard. I'd love to get rid of some of the `interface{}` and casting and of course the deep equals and replace it with generics. Go generics is an option now that we're on 1.20 as a project. I think we have the possibility of introducing some generics from the ground up in libcnb and libpak which might make this all cleaner.

c0d1ngm0nk3y: > I'm just struggling with the risk/benefit on this one.

tbh, I don't see any risk with that one. I kind of see your point, but I think it is important to keep the current version in a good shape and not push everything to the next shiny release. But I see that this might be tempting. But it is your call.

> I wouldn't be opposed to doing something like this in v2.

Then I will open a pr for `v2` instead. But this will mean that the cleanup step will take quite a while because it will take some time until `v2` is consumed.

BTW, is there a ETA for `v2`?

dmikusa: I think the libcnb v2 repo is in pretty good shape. It's possible things could still change, but it's only likely to change with regards to support for the next buildpack API & removing stacks (0.10).

As such, we've been working on v2 for libpak. We're tagging v2 alpha releases so it's easy for folks to start testing them. There's still a high probability for code change, so I wouldn't start porting buildpacks unless you want to be part of the effort in designing and evolving the libpak API, then it's a perfect time to get involved.

It all kind of goes hand-in-hand, but we're doing the same with pipeline-builder & libjvm. We've not started libbs yet.

c0d1ngm0nk3y:
I don't fully understand the reasoning, but I will close this pr for `v1` and #283 is open for `v2`.

PR #273, opened 2023-07-28
adds logic to remove http query params from destination file name

### LINKED ISSUE #274: Avoid including http query parameters in downloaded (http) artifact's filename
## Describe the Enhancement
This request is in the context of Paketo's apache-tomcat [buildpack](https://github.com/paketo-buildpacks/apache-tomcat/tree/main). 
We are trying to use a presigned S3 url to download tomcat external configurations, setting the url as the value for the environment variable `BP_TOMCAT_EXT_CONF_URI`. Internally the buildpack [calls](https://github.com/paketo-buildpacks/apache-tomcat/blob/00745a96587fb53984e7b478d5a24aaf0bf78335/tomcat/base.go#L275) libpak. DependencyCache's `Artifact` function.
This function sets the name of the downloaded artifact based on the value returned by the `filepath.Base(uri)` function [call](https://github.com/paketo-buildpacks/libpak/blob/20229c947f55ede8e5c95eda98c5f604a3961ebe/dependency_cache.go#L185).
However, if we use an http uri with query parameters this call will try to set the filename with the query parameters included.

In our case, the S3 presigned url we are using has very long query parameter values, and when `libpak` tries to save the downloaded file with these parameters included in the filename, it runs into _**file name too long**_ error.

## Possible Solution
We can parse the uri using net/url's `Parse` function and then get the clean uri without the query parameters using the `Path` property. We can also put checks to ensure this is done only for http/https URIs.

## Motivation
This would add support for presigned S3 urls to the Tomcat buildpack to provide external configurations.

We already have a PR drafted, let us know if we can mark it ready for review.

semmet95:
## Summary
This PR adds changes to ensure that downloaded **http** type dependency artifacts are given a filename that does not include the query parameters that might be present in the corresponding URI.
- parse the uri using net/url's Parse function 
- get the clean uri without the query parameters using the Path property
- use this clean uri to get the name for the downloaded artifact

## Use Cases
fixes #274 

## Checklist

### DIFF
--- a/dependency_cache.go
+++ b/dependency_cache.go
@@ -182,7 +182,19 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 			color.New(color.FgYellow, color.Bold).Sprint("Warning:"))
 
 		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
-		artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
+
+		// ensure query parameters are not included in the downloaded file name if the uri is http type
+		downloadUri, err := url.Parse(uri)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse the download uri %s\n%w", uri, err)
+		}
+
+		if(downloadUri.Scheme == "http" || downloadUri.Scheme == "https") {
+			artifact = filepath.Join(d.DownloadPath, filepath.Base(downloadUri.Path))
+		} else {
+			artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
+		}
+
 		if err := d.download(uri, artifact, mods...); err != nil {
 			return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
 		}
--- a/dependency_cache.go
+++ b/dependency_cache.go
@@ -190,7 +190,10 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 		}
 
 		if(downloadUri.Scheme == "http" || downloadUri.Scheme == "https") {
-			artifact = filepath.Join(d.DownloadPath, filepath.Base(downloadUri.Path))
+			hasher := sha256.New()
+			hasher.Write([]byte(uri))
+
+			artifact = filepath.Join(d.DownloadPath, hex.EncodeToString(hasher.Sum(nil)))
 		} else {
 			artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
 		}
--- a/dependency_cache.go
+++ b/dependency_cache.go
@@ -167,7 +167,9 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 		actual   BuildpackDependency
 		artifact string
 		file     string
+		hasher   = sha256.New()
 		uri      = dependency.URI
+		uriSha   string
 	)
 
 	for d, u := range d.Mappings {
@@ -177,26 +179,17 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 		}
 	}
 
+	// downloaded artifact will have the uri sha as its filename
+	hasher.Write([]byte(uri))
+	uriSha = hex.EncodeToString(hasher.Sum(nil))
+
 	if dependency.SHA256 == "" {
 		d.Logger.Headerf("%s Dependency has no SHA256. Skipping cache.",
 			color.New(color.FgYellow, color.Bold).Sprint("Warning:"))
 
 		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
 
-		// ensure query parameters are not included in the downloaded file name if the uri is http type
-		downloadUri, err := url.Parse(uri)
-		if err != nil {
-			return nil, fmt.Errorf("unable to parse the download uri %s\n%w", uri, err)
-		}
-
-		if(downloadUri.Scheme == "http" || downloadUri.Scheme == "https") {
-			hasher := sha256.New()
-			hasher.Write([]byte(uri))
-
-			artifact = filepath.Join(d.DownloadPath, hex.EncodeToString(hasher.Sum(nil)))
-		} else {
-			artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
-		}
+		artifact = filepath.Join(d.DownloadPath, uriSha)
 
 		if err := d.download(uri, artifact, mods...); err != nil {
 			return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
@@ -234,7 +227,7 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 	}
 
 	d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
-	artifact = filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(uri))
+	artifact = filepath.Join(d.DownloadPath, dependency.SHA256, uriSha)
 	if err := d.download(uri, artifact, mods...); err != nil {
 		return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
 	}
--- a/dependency_cache_test.go
+++ b/dependency_cache_test.go
@@ -20,6 +20,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
+	"hash"
 	"io"
 	"net/http"
 	"os"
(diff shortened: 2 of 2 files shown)


dmikusa:
Hi, I took a look. A couple of things on this:

1. It needs test coverage. There might be existing tests covering HTTP and file downloads, but we need to make sure there's one also with query params and confirm that the expected file name is produced.

2. Thinking through test scenarios, I also worry about the case where you have a URL like https://www.example.com/download.php?give-me=something. If we throw out the query params, you could have collisions with downloads and that wouldn't be good. It might make more sense to just sha256 hash the full URL and use that for the file name. I don't like that it creates meaningless file names, but it will ensure a unique file name for the download URL. Is this a change you'd be willing to make?

semmet95:
Thanks for the suggestions @dmikusa 
I'll look into them and update the PR accordingly

semmet95:
I checked just to be sure that the `apache-tomcat` buildpack doesn't need the file name to have the proper extension and even with the sha256 as its name the buildpack managed to download and extract the tomcat configuration tar file.
We are working on the suggested changes now.

semmet95:
@dmikusa pushed the changes. Please let me know if any improvements can be made.

semmet95:
@paketo-bot-reviewer requesting review

pranavek:
@dmikusa The changes are implemented as per your suggestions. Could you take a look at it? 

NB: Some Github workflows are in `Waiting for status` state.

dependency_cache_test.go
```diff
@@ -317,6 +319,21 @@ func testDependencyCache(t *testing.T, context spec.G, it spec.S) {
...
+			server.AppendHandlers(ghttp.RespondWith(http.StatusOK, "alternate-fixture"))
+
+			hasher := sha256.New()
+			hasher.Write([]byte(dependency.URI))
+
+			a, err := dependencyCache.Artifact(dependency)
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(io.ReadAll(a)).To(Equal([]byte("alternate-fixture")))
+			Expect(filepath.Base(a.Name())).To(Equal(hex.EncodeToString(hasher.Sum(nil))))
```
anthonydahanne, line 334: instead of re doing what you do in the implementation (`hasher` creation), you could have just compared the test filename with the String representation of the hash?

anthonydahanne:
I've allowed the action to run and I rebased on main.

Looks good to me, @dmikusa WDYT?

dependency_cache.go
```diff
@@ -182,7 +182,22 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
...
 		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
-		artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
+
+		// ensure query parameters are not included in the downloaded file name if the uri is http type
+		downloadUri, err := url.Parse(uri)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse the download uri %s\n%w", uri, err)
+		}
+
+		if(downloadUri.Scheme == "http" || downloadUri.Scheme == "https") {
```
dmikusa, line 192: Thinking about this. Should we have an if block here that checks if there are query params? Logically, I believe we only need to do this if there are query params.

Conversely, I'd say we just hash things across the board, but I'd want to check if that would have any side effects before doing that.

anthonydahanne: 👍  we could consider it's always http/https and remove this check

dmikusa: I looked over the code again and I think it's safe to just use the hasher all the time. That means you don't need to parse the URL, you can just feed the full URL as a string into the hasher and use the hash as the file name.

That said, you also need to do this at https://github.com/paketo-buildpacks/libpak/blob/6a1c758bae92ef4ff1bca7e40c71a65b4140ec74/dependency_cache.go#L236-L240

The code you've modified here is for the case where the dependency has no sha256 in the metadata. If there is a sha256 in the metadata, then you'll skip this whole block and it will download the file on lines 236-240 down below.

dependency_cache_test.go
anthonydahanne: yeah, this is what I usually do; since the shasum is not gonna change from one run to another, then instead of repeating what you did in the implementation, you could just write your test with this assumption:  

> I know the shasum of `http://example.com/test-path?param1=value1&param2=value2` is `ffff` so I assert the impl. will return `ffff` as the filename

That way, you decouple your knowledge of the underlying code, not caring how the impl. code was written.

I hope that makes sense

semmet95: Ah ok, it makes sense, but when I tried comparing directly with the shasum, but it seems the `httptest` library is [randomly selecting ](https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/net/http/httptest/server.go;l=24)an available port for the mocked server and that in turn keeps changing the dependency uri, so I couldn't find a way to hardcode the shasum in the comparison.

anthonydahanne: ah! that's a good point, the URL is dynamic, I had not noticed.

hum, well I guess you can not predict the shasum then 🤷

In that case, I believe you already chose for your test the best outcome!

All good!

semmet95: haha, purely accidently
hadn't noticed only until now that the url is not static

dependency_cache.go
semmet95: got it, I'll start working on the changes

semmet95: @dmikusa updated. URI sha will now be used as the downloaded artifact's filename regardless of the value of dependency.SHA256

dependency_cache_test.go
```diff
@@ -317,6 +319,21 @@ func testDependencyCache(t *testing.T, context spec.G, it spec.S) {
 			Expect(io.ReadAll(a)).To(Equal([]byte("alternate-fixture")))
 		})
 
+		it("sets downloaded file name to uri's sha256 with empty SHA256 and query parameters in the uri", func() {
```
semmet95, line 322: - [ ] add a new test for a normal uri without query params and with non empty SHA256

dmikusa (approved):

dmikusa:
Thanks for this PR & sticking with it. If you're interested, we'll need to port this to the `release-2.x` branch as well.

c0d1ngm0nk3y:
Some feedback: This change was breaking. I wonder why this was brought into `v1`. Just admitting it in the release notes is not the best user experience. At least I do not read the release notes carefully for each minor bump.

dmikusa:
@c0d1ngm0nk3y Apologizes. We try to screen out any breaking changes, but this one slipped through. We wanted to get this into 1.x because it impacted users. In hindsight, we'd have held it for 2.x.

c0d1ngm0nk3y:
@dmikusa But doesn't make it clear that some integration test was missing. I mean it was not the case that one very specific edge case was broken, but offline buildpacks, didn't work in general, right? It feels to me that this should be tested somehow.

What wonders me more is not the fact that this was overseen - mistakes happen, that normal imho. But rather the reaction. Knowing that this is breaking but still keep the code and not reverting the change was strange to me.

 My POV: I discovered that the integration test for one of our use cases broke. After some debugging, I found out that it is nothing we changed, but that offline buildpacks do not work in general. When checking the release notes of `create-package`, I saw that this problem is known and there there is even a notice buried in the release notes of a minor patch. The experience was not the best and it felt a bit unnecessary to me.

dmikusa:
> But doesn't make it clear that some integration test was missing.

100%. We have stories to add integration testing to the composite buildpacks, but the work is incomplete. We'd be happy to work with you on that, if it's something you'd like to contribute.

> What wonders me more is not the fact that this was overseen - mistakes happen, that normal imho. But rather the reaction. Knowing that this is breaking but still keep the code and not reverting the change was strange to me.

What you're seeing is absolute transparency and us working through this problem in real-time. We could have taken the bug report, sat on this for a week or so while we investigated the full extent of the problem, made a decision, and then finally communicated that to the community. My approach for the community is to communicate early and often. There are some drawbacks to that because we're working through the issue in real-time, but I think the benefit of being transparent with everything outweighs the approach of keeping things private until they've been fully sorted out.

That said. here's the latest update. We're reevaluating this. Another bug case has popped up. Previously, the decision was made to try and keep moving forward because we wanted to be able to retain the original fix that #273 provided. Additional failure scenarios have popped up. We're presently discussing those, the potential impact on the user base, and the work required to resolve them. We'll keep folks updated when a course of action has been decided upon.

The recommendation is still to continue using the older version of create-package and libpak. This should avoid issues because both are using the pre-#273 code.

pivotal-david-osullivan:
Having investigated the bug raised in #288 as well as further failures seen, we've decided to revert the change in this PR to minimise impact and avoid introducing other issues by attempting to patch the new logic in the short term.

Hopefully the problem scenario faced in #274 can be resolved in v2 of Libpak (#287)

PR #312, opened 2024-02-02
Obfuscate Credentials in Logs

bitgully:
## Summary

When using HTTP basic authentication credentials in binding URIs, this change obfuscates the username and password in the log entries.

## Use Cases

When adding a binding for an artifact hosted in a private registry, it can be necessary to provide HTTP basic authentication credentials with the URI. This already works fine for downloading the desired dependency. However, when doing so, the username and password that are part of the URI, get printed to stdout in plain text at the moment. This PR should solve this issue by obfuscating the credentials.
To make sure this change has no adverse side effects in case the provided URI is not well-formed, this implementation uses a regular expression instead of trying to parse the string to a "net/url" object.

## Checklist

### DIFF
--- a/dependency_cache.go
+++ b/dependency_cache.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -181,10 +182,10 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 		d.Logger.Headerf("%s Dependency has no SHA256. Skipping cache.",
 			color.New(color.FgYellow, color.Bold).Sprint("Warning:"))
 
-		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
+		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), sanitizeUri(uri))
 		artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
 		if err := d.download(uri, artifact, mods...); err != nil {
-			return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
+			return nil, fmt.Errorf("unable to download %s\n%w", sanitizeUri(uri), err)
 		}
 
 		return os.Open(artifact)
@@ -218,10 +219,10 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 		return os.Open(filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(uri)))
 	}
 
-	d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
+	d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), sanitizeUri(uri))
 	artifact = filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(uri))
 	if err := d.download(uri, artifact, mods...); err != nil {
-		return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
+		return nil, fmt.Errorf("unable to download %s\n%w", sanitizeUri(uri), err)
 	}
 
 	d.Logger.Body("Verifying checksum")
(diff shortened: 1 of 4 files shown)


dependency_cache.go
```diff
@@ -334,12 +335,18 @@ func (d DependencyCache) downloadHttp(uri string, destination string, mods ...Re
...
 
 	return nil
 }
 
+func sanitizeUri(uri string) string {
+	uriCreds := regexp.MustCompile(`:\/\/.*:.*@`)
+	uriCredsRepl := "://***:***@"
+	return uriCreds.ReplaceAllString(uri, uriCredsRepl)
+}
+
```
dmikusa, line 349: I think my preference would be to parse the URI using `url.Parse` and then call the `Redacted` method on it. https://pkg.go.dev/net/url#URL.Redacted. That will redact the password. I feel like that is probably sufficient, but if you feel strongly that we should redact both the user and password, then I would suggest implementing it in a similar way.

You can parse the URL and then replace the user/password part in the parsed component, then write it out. Something like this: https://go.dev/play/p/cmdsJ7pyB28

The tricky part is that parsing can fail & how should we handle that? 

I think my thought would be to attempt to parse it. If the parse fails, then log the URL & error at *debug* level, then have the redact/sanitize method return a string like "URL has been redacted (see DEBUG log level)" and just not display any of it.

bitgully: I agree this would be a cleaner solution. It would also be in line with how the http client itself behaves when failing to fetch a URL (obfuscates password only). I will get at it soon.

dmikusa: Thanks for the update!

dependency_cache.go
```diff
@@ -177,14 +178,20 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
...
 	if dependency.SHA256 == "" {
 		d.Logger.Headerf("%s Dependency has no SHA256. Skipping cache.",
 			color.New(color.FgYellow, color.Bold).Sprint("Warning:"))
 
-		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
-		artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
-		if err := d.download(uri, artifact, mods...); err != nil {
-			return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
+		d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), urlP.Redacted())
+		artifact = filepath.Join(d.DownloadPath, filepath.Base(urlP.Path))
```
dmikusa, line 192:
```suggestion
		artifact = filepath.Join(d.DownloadPath, filepath.Base(uri))
```

I'd like to change this back to use the base of the URI. There is technically a difference if you have query parameters on the URI and I'm not sure if it would impact anyone, but I'd rather not chance it. I don't think this affects the redaction of values, so I think this should be OK, but please let me know if you have concerns.

dependency_cache.go
```diff
@@ -215,13 +222,13 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
...
-		return os.Open(filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(uri)))
+		return os.Open(filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(urlP.Path)))
 	}
 
-	d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), uri)
-	artifact = filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(uri))
-	if err := d.download(uri, artifact, mods...); err != nil {
-		return nil, fmt.Errorf("unable to download %s\n%w", uri, err)
+	d.Logger.Bodyf("%s from %s", color.YellowString("Downloading"), urlP.Redacted())
+	artifact = filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(urlP.Path))
```
dmikusa, line 229:
```suggestion
	artifact = filepath.Join(d.DownloadPath, dependency.SHA256, filepath.Base(uri))
```

Same comment here as above.

dmikusa (changes requested):
Two small changes. If those are OK, let me know and I'll merge everything. Thanks!

bitgully:
Sure, they are okay. Thanks for the review.

PR #315, opened 2024-02-23
Allow for overriding default dependency source

bitgully:
## Summary
Using the introduced environment variable BP_DEPENDENCY_SOURCE_OVERRIDE we can override the default source (scheme://host:port) of dependencies. Excluding those explicitly defined in bindings.

## Use Cases
In (air-gapped) corporate networks, it is common to work with mirror registries. Dependencies cannot be pulled directly from the public internet but must run through the private mirror for security and performance/resilience reasons.
In such cases, we would have to manually define several individual dependency-mapping bindings for each dependency version. It is more practical to set the mirror's address once for all dependencies.

### Sample
Setting variable BP_DEPENDENCY_SOURCE_OVERRIDE=https://mirror.acme.com would lead to this URI replacement:
Before: https://github.com/bell-sw/Liberica/releases/download/11.0.8+10/bellsoft-jre11.0.8+10-linux-amd64.tar.gz
After: https://mirror.acme.com/bell-sw/Liberica/releases/download/11.0.8+10/bellsoft-jre11.0.8+10-linux-amd64.tar.gz

## Checklist


dmikusa:
This is something that we've resisted doing in the past. Dependency mappings would be the suggested way to do this instead, or build your own buildpack images and then you can point the dependencies to wherever you want.

Security and provenance have been the primary motivators for not doing this previously. What you've proposed is a little different though, only being able to modify the `scheme://host:port`. Off the top of my head, I'd like this more if it were just allowing you to change the `host:port`, because everything should always be HTTPS. Would that still work for you?

I appreciate the PR. I need to think about this some more and see what the other maintainers think as well.

Thanks!

anthonydahanne (commented):
LGTM

bitgully:
Sure, HTTPS only would work for me too. Overriding the scheme was just meant for backwards compatibility or in case anyone has a use case without TLS (maybe for performance reasons).
Let me know your preference when you are ready, and I will adapt it accordingly.

dmikusa:
Excellent. Yes, please update to require HTTPS.

Other thoughts:

1. I hate naming things cause it's hard and I don't like being picky about names cause it's subject but I think we should discuss the env variable name a bit. The term "source" was a little confusing to me. My preference would be to try to name it using the same terminology that we're using now. We have "dependency mappings" to remap a particular dependency, so maybe we have a "dependency host mapping"? or a "dependency domain mapping"? or a "dependency server mapping"? To indicate you're remapping an entire host/domain/server. What do others think about this?

2. I'm curious about using an env variable versus a binding, like with a dependency mapping? Did you have a specific reason for env variable? Just thinking through things in my head. If we used a binding instead of an env variable, then the process would be more consistent with the existing dependency mapping process. It would also facilitate using embedded HTTP basic credentials, which is something that would be nice to support since you wouldn't want to put those in an env variable.

3. HTTP basic credentials. I feel like this is something people will want/need. If you don't want to include this now, I'm fine with that, but it's something we should think through so that we can support it in the future.

4. I think we should somehow indicate a change like this has happened to the user. Definitely at build time. Possibly in the SBOM information or possibly as a label on the generated container? 🤔 but I would prefer in the SBOM. I just want to make sure that it's clear how the container was created and that provenance is clear as well.

dmikusa:
My thought in regards to the SBOM is that we've now got different potential sources for the files. You can't tamper with them, because of the SHA256 hash, but it would still be helpful to know exactly where the files were sourced from. It's probably something we *should* do with dependency mapping, but I don't believe we do at the moment.

I'm also OK with spinning that out into another issue, but just so that we're thinking it through and doing things in a way where it is possible.

Oh, the other thing I was thinking about. We need to support `file://` as well as `https://`. A big driver of dependency mapping is to switch from external to local URLs (typically added via a volume mount). We should support that here, because it'll make using local files a lot simpler. If that's a lot more work, again, I'm OK spinning that out into a separate issue.

bitgully:
@loewenstein: A hint, that the original URI was overridden, is printed to the build log [here](https://github.com/paketo-buildpacks/libpak/blob/7e14da4069a2d6f75a633f009b590d5eadff1345/dependency_cache.go#L376).

@dmikusa:
1. True, naming is hard. I also thought of things like: "dependency registry", "dependency mirror", "dependency mapping proxy", "dependency mapping location". But I'm happy with any other version really.
2. The reason I chose an env variable was that I mostly work in Kubernetes environments, where the env variable can be injected into the build container from a Kubernetes Secret and therefore would only exist inside the container. RBAC rules define who can view/edit such Secrets. I agree, when not running on a container orchestration platform, using an env variable might not be feasible. But having embedded HTTP credentials saved inside a binding file seemed less than ideal too to me.
3. The HTTP basic credentials would in general be supported in this change [here](https://github.com/paketo-buildpacks/libpak/blob/7e14da4069a2d6f75a633f009b590d5eadff1345/dependency_cache.go#L380).
4. Regarding the SBOM: I simply considered that mirror more like an external cache and didn't see it as an alternative location since artefacts would only be proxied through this. However, one could also use this override mechanism to simply point to any other server that actually isn't set up as a mirror/proxy. As the same applies to any dependency mapping, it would make sense to move this into a separate issue though.

bitgully:
I have decided to use "BP_DEPENDENCY_MIRROR" as a name now and limited the support for the https:// and file:// schemes only (no more http:// allowed).
I'm planning to add the option of setting a mirror using a binding instead of an env variable at a later stage as well. But I'm tied up in a few other projects at the moment.
But I have introduced an additional feature "BP_DEPENDENCY_MIRROR_PRESERVE_HOST" too. This can be necessary for use cases, where it isn't sufficient to just replace the hostname but also preserve the original hostname in the path on the mirror. Depending on the mirror's setup, this can be vital to ensure uniqueness of the (mirrored/proxied) sources.

dependency_cache.go
```diff
@@ -370,19 +381,27 @@ func (DependencyCache) verify(path string, expected string) error {
...
 
-func (d DependencyCache) overrideDependencySource(urlD *url.URL) {
-	dependencySourceOverride := sherpa.GetEnvWithDefault("BP_DEPENDENCY_SOURCE_OVERRIDE", "")
-	if dependencySourceOverride != "" {
-		d.Logger.Infof("variable BP_DEPENDENCY_SOURCE_OVERRIDE found. overriding download uri.")
-		urlOverride, err := url.ParseRequestURI(dependencySourceOverride)
-		if err == nil {
+func (d DependencyCache) setDependencyMirror(urlD *url.URL) {
+	dependencyMirror := sherpa.GetEnvWithDefault("BP_DEPENDENCY_MIRROR", "")
+	dependencyMirrorPreserveHost := sherpa.GetEnvWithDefault("BP_DEPENDENCY_MIRROR_PRESERVE_HOST", "false")
```
dmikusa, line 386: These settings should be moved to the DependencyCache struct.

I would suggest you then update `NewDependencyCache` such that it pulls from env variables (i.e. default behavior). That will mean it only pulls from the env variables once, instead of every download.

I think it will make testing easier too. You won't need to set the env variables, you can just make a new struct and set the config values.

dependency_cache.go
```diff
@@ -187,7 +188,7 @@ func (d *DependencyCache) Artifact(dependency BuildpackDependency, mods ...Reque
 	}
 
 	if !isBinding {
-		d.overrideDependencySource(urlP)
+		d.setDependencyMirror(urlP)
```
dmikusa, line 191: I think we should log a warning if there are both dependency mapping bindings and the dependency mirror setting being invoked at the same time.

Something like: `Both BP_DEPENDENCY_MIRROR and dependency bindings are present. BP_DEPENDENCY_MIRROR is being ignored. Please remove dependency map bindings if you wish to use BP_DEPENDENCY_MIRROR.`.

dependency_cache.go
```diff
@@ -370,19 +381,27 @@ func (DependencyCache) verify(path string, expected string) error {
...
+		if strings.EqualFold(dependencyMirrorPreserveHost, "true") && urlD.Hostname() != "" {
+			originalHost = "/" + urlD.Hostname()
+		}
+
+		if strings.EqualFold(urlOverride.Scheme, "https") || strings.EqualFold(urlOverride.Scheme, "file") {
 			urlD.Scheme = urlOverride.Scheme
 			urlD.User = urlOverride.User
 			urlD.Host = urlOverride.Host
-			urlD.Path = urlOverride.Path + urlD.Path
+			urlD.Path = urlOverride.Path + originalHost + urlD.Path
```
dmikusa, line 401: I'm not sure I understand what's going on here and the `BP_DEPENDENCY_MIRROR_PRESERVE_HOST`. Can you expand on the use case for that and how it works? Is that something specific that some servers require? Thanks

dependency_cache.go
dmikusa: Also, I would suggest `sherpa.ResolveBool` or `sherpa.ResolveBoolErr` here instead, that way you have a bool to work with and you don't need string comparisons.

https://github.com/paketo-buildpacks/libpak/blob/main/sherpa/env_var.go#L57

dependency_cache.go
bitgully: The thing is, paths and artifact names are unique only within the scope of one registry/hostname.

Here is a real world example where the same artifact is made available in different registries:
https://repo1.maven.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz
https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz

If our local mirror acts as a proxy for both "repo1.maven.org" as well as for "repo.maven.apache.org", substituting the hostname only might lead to unknown results because the mirror wouldn't know what upstream registry to target.
By setting BP_DEPENDENCY_MIRROR=https://local-mirror.acme.com and BP_DEPENDENCY_MIRROR_PRESERVE_HOST=true, we could still ensure uniqueness.
Here are the overridden URIs for this case:
https://local-mirror.acme.com/repo1.maven.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz
https://local-mirror.acme.com/repo.maven.apache.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz

Usually, the combination of path + filename itself is unique across registries and hostname preservation should not be needed. I just thought, one day, two different artifacts might be hosted under the same path/filename by chance.

dmikusa: Ah, ok. I see what you're saying. Do you think we should just genericize this and have a `BP_DEPENDENCY_MIRROR_PREFIX` or something like that?

I've been wondering how much we should allow people to change the path of the URL, but this makes sense. I think the name mirror implies that you'll have the same path suffix, but you definitely might need a different prefix.

bitgully: The prefixing option is provided as part of the "BP_DEPENDENCY_MIRROR" at the moment.

Let's assume the original URI is:
https://repo1.maven.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz

The current implementation already allows setting a prefix. You can either set scheme://hostname only, or include a prefix using scheme://hostname/prefix. But this prefix part is static.

Scenario A:
With BP_DEPENDENCY_MIRROR=https://local-mirror.acme.com
Would result in: https://local-mirror.acme.com/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz

Scenario B:
With BP_DEPENDENCY_MIRROR=https://local-mirror.acme.com/custom-prefix
Would result in: https://local-mirror.acme.com/custom-prefix/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz

Now, the BP_DEPENDENCY_MIRROR_PRESERVE_HOST=true can be used additionally in both scenarios. The difference to the prefix is, that the hostname part in the path is dynamic. This part changes according to the original download source of each dependency.

If we'd add BP_DEPENDENCY_MIRROR_PRESERVE_HOST=true in both scenarios from above, the results would be:
Scenario A:
https://local-mirror.acme.com/repo1.maven.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz
Scenario B:
https://local-mirror.acme.com/custom-prefix/repo1.maven.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz

I believe, the current solution leaves room for pretty much any customization.
Would you prefer to have an additional variable "BP_DEPENDENCY_MIRROR_PREFIX", instead of just appending it to the existing "BP_DEPENDENCY_MIRROR" as the path (as in scenario B).

dmikusa: OK, I see. I was thinking of `BP_DEPENDENCY_MIRROR` as just the protocol/host/port, but if we're allowing Path to be included and we treat that as a Path Prefix, I'm OK with that. I think we'll just need to document how that works so folks know it's a prefix that's added not the full path. Ultimately, it's less configuration, and I see how it's different. 

I'm just thinking out loud about ways we could reduce the configuration. Less configuration is usually better.

One other idea. What about just having a placeholder in the Path prefix or the URL like `{originalHost}` or something where we do a string replacement if that token is included?

Ex: `BP_DEPENDENCY_MIRROR=https://local-mirror.acme.com/{originalHost}/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz` would result in `https://local-mirror.acme.com/repo1.maven.org/maven2/org/apache/maven/apache-maven/4.0.0-alpha-9/apache-maven-4.0.0-alpha-9-bin.tar.gz`

This could position us well for future requests, as I could see getting requests for other dynamic changes to the URL path.

bitgully:
I have reworked this now. The changes are:
- Placeholder {originalHost} can be used in arbitrary locations of the mirror path, eliminating BP_DEPENDENCY_MIRROR_PRESERVE_HOST.
- Mirror URIs can also be set using a binding of type 'dependency-mirror' and the filename 'uri'.
- A warning is logged if a dependency mapping is found when a mirror is defined at the same time.

dmikusa:
Thank you! Sorry for the delay. I will get back to this today or tomorrow.

dmikusa (approved):

dmikusa:
I think this came out really awesome. Thanks so much for sticking through the review and incorporating feedback.

@anthonydahanne Are you 👍 also? If so, I'll merge.

anthonydahanne (approved):
thanks for your contribution @bitgully ! Looks great!

dmikusa:
Moving this to the RFC.

PR #322, opened 2024-04-02
Allow hostname specific dependency mirrors

bitgully:
## Summary
Enhances the use of dependency mirrors introduced in #315.
Changes have been applied according to the conversation in https://github.com/paketo-buildpacks/rfcs/pull/302 ([Readable](https://github.com/paketo-buildpacks/rfcs/blob/dep-mirrors/text/0000-dependency-mirrors.md)).

This change allows for setting separate mirror locations for each hostname in the dependencies' original URIs.
A default mirror can still be used for all other dependencies.

## Use Cases
This is required in networks where not all dependencies should be pulled from the same mirror location.

## Checklist

### DIFF
--- a/dependency_cache.go
+++ b/dependency_cache.go
@@ -154,7 +154,7 @@ func customizeHttpClientTimeouts() (*HttpClientTimeouts, error) {
 }
 
 // Returns the URI of a dependency mirror (optional).
-// Such mirror location can be defined in a binding of type 'dependency-mirror' with filename 'uri'
+// Such mirror location can be defined in a binding of type 'dependency-mirror' with filename 'default'
 // or using the environment variable 'BP_DEPENDENCY_MIRROR'. The latter takes precedence in case both are found.
 func getDependencyMirror(bindings libcnb.Bindings) (string, error) {
 	dependencyMirror := sherpa.GetEnvWithDefault("BP_DEPENDENCY_MIRROR", "")
@@ -162,8 +162,8 @@ func getDependencyMirror(bindings libcnb.Bindings) (string, error) {
 	if dependencyMirror == "" {
 		dependencyMirrorBindings, err := filterBindingsByType(bindings, "dependency-mirror")
 		if err == nil {
-			// Use the content of the file named "uri" as the mirror's URI.
-			dependencyMirror = dependencyMirrorBindings["uri"]
+			// Use the content of the file named "default" as the mirror's URI.
+			dependencyMirror = dependencyMirrorBindings["default"]
 		} else {
 			return "", err
 		}
--- a/dependency_cache.go
+++ b/dependency_cache.go
@@ -69,28 +69,25 @@ type DependencyCache struct {
 	// httpClientTimeouts contains the timeout values used by HTTP client
 	HttpClientTimeouts HttpClientTimeouts
 
-	// Alternative source used for downloading dependencies.
-	DependencyMirror string
+	// Alternative sources used for downloading dependencies.
+	DependencyMirrors map[string]string
 }
 
 // NewDependencyCache creates a new instance setting the default cache path (<BUILDPACK_PATH>/dependencies) and user
 // agent (<BUILDPACK_ID>/<BUILDPACK_VERSION>).
 // Mappings will be read from any libcnb.Binding in the context with type "dependency-mappings".
 //
-// In some air-gapped environments, dependencies might not be download directly but need to be pulled from a local mirror registry.
-// In such cases, an alternative URI can either be provided as environment variable "BP_DEPENDENCY_MIRROR", or by a binding of type "dependency-mirror"
-// where a file named "uri" holds the desired location.
-// The two schemes https:// and file:// are supported in mirror URIs where the expected format is (optional parts in "[]"):
-// <scheme>://[<username>:<password>@]<hostname>[:<port>][/<prefix>]
-// The optional path part of the provided URI is used as a prefix that might be necessary in some setups.
-// This (prefix) path may also include a placeholder of "{originalHost}" at any level (in sub-paths or at top-level) and is replaced with the
-// hostname of the original download URI at build time. A sample mirror URI might look like this: https://local-mirror.example.com/buildpacks-dependencies/{originalHost}
+// In some environments, many dependencies might need to be downloaded from a (local) mirror registry or filesystem.
+// Such alternative locations can be configured using bindings of type "dependency-mirror", avoiding too many "dependency-mapping" bindings.
+// Environment variables named "BP_DEPENDENCY_MIRROR" (default) or "BP_DEPENDENCY_MIRROR_<HOSTNAME>" (hostname-specific mirror)
+// can also be used for the same purpose.
 func NewDependencyCache(context libcnb.BuildContext) (DependencyCache, error) {
 	cache := DependencyCache{
-		CachePath:    filepath.Join(context.Buildpack.Path, "dependencies"),
-		DownloadPath: os.TempDir(),
-		UserAgent:    fmt.Sprintf("%s/%s", context.Buildpack.Info.ID, context.Buildpack.Info.Version),
-		Mappings:     map[string]string{},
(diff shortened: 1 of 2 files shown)


dependency_cache.go
```diff
@@ -181,7 +202,7 @@ func filterBindingsByType(bindings libcnb.Bindings, bindingType string) (map[str
 				if _, ok := filteredBindings[key]; ok {
 					return nil, fmt.Errorf("multiple %s bindings found with duplicate keys %s", binding.Type, key)
 				}
-				filteredBindings[key] = value
+				filteredBindings[strings.ToLower(key)] = value
```
dmikusa, line 205: Two things about this:

1. Small thing. I think we'd need to `strings.ToLower(key)` the check on line 202 as well. Otherwise, it would use a potentially non-lowercase key to look in the map which we're then only putting lowercase keys into.

2. I worry about changing the behavior here. Previously, this was filtering case sensitively. We could impact users with existing bindings. Is there a strong need to change how this works? I think my preference would be to not change this here, unless there's a strong reason to. If nothing else, for backward compatibility.

dependency_cache.go
```diff
@@ -153,22 +150,46 @@ func customizeHttpClientTimeouts() (*HttpClientTimeouts, error) {
...
 	}
-	return dependencyMirror, nil
+	return mirrors
+}
+
+// Takes an encoded hostname (from env key) and returns the decoded version in lower case.
+// Replaces double underscores (__) with one dash (-) and single underscores (_) with one period (.).
+func decodeHostnameEnv(encodedHostname string) string {
+	var decodedHostname string
+	if encodedHostname == "" {
```
dmikusa, line 187: I think we should check and forbid someone from entering "default" as a hostname. If you did `BP_DEPENDENCY_MIRROR_DEFAULT`, I think, that would result in `encodedHostname` being `default`. I'm not sure that would be a problem exactly, but it would be defensive of us to skip that and just log a warning like "You have a mirror host name of `default`, which is prohibited. It has been ignored. Use `BP_DEPENDENCY_MIRROR` to set the default mirror.

dependency_cache.go
```diff
@@ -153,22 +150,46 @@ func customizeHttpClientTimeouts() (*HttpClientTimeouts, error) {
...
+		decodedHostname = "default"
+	} else {
+		decodedHostname = strings.ReplaceAll(strings.ReplaceAll(encodedHostname, "__", "-"), "_", ".")
```
dmikusa, line 190: Should check for characters outside of the ones that are allowed? Like `-`'s or `.`'s or even lowercase letters before we convert. I don't think we need to fail if there are invalid characters (although maybe if it's characters outside of what's allowed in a hostname?), because somehow those made it through the shell to us. If they make it through that's OK, but shells are weird and maybe this works on a dev's laptop but then not in CI. That's just a weird problem to troubleshoot. If we output some sort of warning in this situation, it would be visible and would train users to do it the way that should consistently work.

We could log a warning like "You have invalid characters in your mirror host environment variable. These will be allowed but for best results across different shells, you should replace `.` characters with `_` characters and `-` characters with `__`, and use all upper case letters. The buildpack will convert these back before using the mirror."

dependency_cache.go
bitgully: Yes, you are right about line 202; this needs to be added.
The switch to case insensitivity is caused by the requirement to override dependency-mirror bindings with those set in env variables (line 163). If the map was initially filled with case sensitive keys (like before), we would have to iterate over those again afterwards instead of accessing the map entry directly.

Regarding the backwards compatibility:
I believe, this code was used only on bindings of type "dependency-mapping" so far, which work with SHA256 checksums as keys. Since these are only hex representations, they can't be considered case sensitive. The only problem I can think of is, if someone didn't put the SHA256 in lowercase in the buildpacks.toml. But this would already fail at verification in line 310.

dependency_cache.go
```diff
@@ -153,22 +150,46 @@ func customizeHttpClientTimeouts() (*HttpClientTimeouts, error) {
...
+	}
+	return dependencyMirrors, nil
+}
+
+// Returns a key/value map of all dependency mirrors set in environment variables.
+func getDependencyMirrorsFromEnv() map[string]string {
+	mirrors := map[string]string{}
+	envs := os.Environ()
+	for _, env := range envs {
+		envPair := strings.SplitN(env, "=", 2)
```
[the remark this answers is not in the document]
bitgully: I think on unix systems, we wouldn't need the additional check. But [this comment on line 63
](https://cs.opensource.google/go/go/+/refs/tags/go1.22.2:src/syscall/env_windows.go;drc=e4ed92a355cebc399dc34d33a556f656fa5c7690;l=63) made me a little nervous now, so I've added it.

dependency_cache.go
bitgully: Has been added.

dependency_cache.go
bitgully: Has been added.

bitgully:
@dmikusa Would you agree with my responses or do you think this needs more attention?
Any specific ideas how we should go about adding this to packit?

loewenstein:
> Any specific ideas how we should go about adding this to packit?

I think that is already in progress with https://github.com/paketo-buildpacks/packit/pull/563

dependency_cache.go
```diff
@@ -88,6 +88,7 @@ func NewDependencyCache(context libcnb.BuildContext) (DependencyCache, error) {
 		UserAgent:         fmt.Sprintf("%s/%s", context.Buildpack.Info.ID, context.Buildpack.Info.Version),
 		Mappings:          map[string]string{},
 		DependencyMirrors: map[string]string{},
+		Logger:            bard.NewLogger(os.Stdout),
```
dmikusa, line 91: Did you add this cause you were having issues with the tests?

The rationale is lost to time, but the Logger is usually set later by the caller after the dependency cache is created. See as [an example](https://github.com/paketo-buildpacks/spring-boot/blob/8501b4fe80f8f6f5a35c36de9a9c3c2b2f06c4d7/boot/build.go#L80-L84). We, unfortunately, can't change this cause it would break the API, but we will fix this in v2.

I ask cause I'd like to keep following that pattern (even though it's not great, consistency helps). If we set the logger here, what you've got will probably be OK (or in most cases overridden because of the pattern I explained), but I'd rather not push it. Then in v2, we'll clean all this up.

dependency_cache.go
dmikusa: > Regarding the backwards compatibility:

That's a good point and you're right. The tooling we use should ensure that all the hashes are in the same case too. I'm OK with changing this.

dmikusa:
Sorry for the delay. I think this is looking good. Just one last question for you.

Thanks again for your work on this!

dependency_cache.go
bitgully: The only reason I added this initialization was to print the warnings you suggested earlier in lines [168](https://github.com/bitgully/libpak/blob/1052fea6ed10479549808574f5f4ec32695de92d/dependency_cache.go#L168) and [181](https://github.com/bitgully/libpak/blob/1052fea6ed10479549808574f5f4ec32695de92d/dependency_cache.go#L181). Since these checks are happening during NewDependencyCache() the usual pattern of setting a logger afterwards is not working. Should we create a bard logger in setDependencyMirrors() instead? I'm happy for any suggestions here.

dmikusa: Oh, ok. In that case, we don't have a choice. Let's just leave it this way. I don't think it will impact anything, I'm just going to add a comment so we don't forget why it's this way. Thanks!

dmikusa, line 91:
```suggestion
		// We create the logger here because the initialization process may log some warnings that should be visible to users.
		// This goes against the usual pattern, which has the user supply the Logger after initialization.
		// There's no choice though, if we want the warning messages to be visible to users. We should clean this up in v2.
		Logger:            bard.NewLogger(os.Stdout),
```

dmikusa (approved):

Issue #326, opened 2024-04-10
Modify create-package to bundle dependencies in a multi-arch offline image


dmikusa:
## Expected Behavior

When you run `create-package` there is the `--include-dependencies` option which downloads and caches all of the deps into the resulting image. This presently assumes there's one image. With the introduction of multi-arch images, the behavior needs to change to support that.

## Possible Solution

1. Download all of the images.
2. Bundle only dependencies specific to a particular architecture of the image of that architecture.

For example, the folder structure for multi-arch is `linux/<arch>/...` so it would make `linux/amd64/dependencies` and `linux/arm64/dependencies` and put the images specific to each architecture in those directories.

In libpak, we call `cache.Artifact` for each artifact, but the cache controls where the files are written so this will likely require setting up two separate caches and then calling the correct cache per artifact (based on its type), or modifying the cache itself to be architecture aware. If the cache itself stored files under sub folders with the architecture name, then you could easily move parts of the cache into the `linux/<arch>/dependencies` directory after caching everything.

I do think that we should wait on this until we switch to using the `arch` metadata field in libpak.

## Motivations

Multi-arch support.

dmikusa:
We have full multi-arch support, so I believe this one is done as well.

PR #331, opened 2024-04-16
Adds `--target-arch` flag

### LINKED ISSUE #330: Add create-package flag so that it outputs a single arch
## Describe the Enhancement

Add a flag to `create-package` that takes in the expected target and then output the buildpack format for a single arch (i.e. older format). In addition, filter the metadata dependencies and only include those for the specified arch.

## Motivation

Not every user will want to build multi-arch images.

dmikusa:
## Summary

Adds a `--target-arch` flag which causes the tool to only create a build pack for the target architecture. By default, it builds all architectures. It still supports the old format of buildpack which does not include arch information in the include-files field.

## Use Cases

Resolves #330 

I'm moving:

> In addition, filter the metadata dependencies and only include those for the specified arch.

into #326 because it'll be easier to add there, cause it's not doing dependencies for multi-arch yet.

## Checklist

### DIFF
--- a/carton/package.go
+++ b/carton/package.go
@@ -35,6 +35,8 @@ import (
 	"github.com/paketo-buildpacks/libpak/internal"
 )
 
+const DefaultTargetArch = "all"
+
 // Package is an object that contains the configuration for building a package.
 type Package struct {
 
@@ -58,6 +60,9 @@ type Package struct {
 
 	// Version is a version to substitute into an existing buildpack.toml.
 	Version string
+
+	// TargetArch is the target architecture to package. Default is "all".
+	TargetArch string
 }
 
 // Create creates a package.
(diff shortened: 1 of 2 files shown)


anthonydahanne (approved):

dmikusa:
@braunsonm - Yes. For now, it will download everything that is in the metadata in buildpack.toml. You can filter dependencies with create-package, but presently it only works by id and by version, not by arch. That's what's going to be added as part of #326.

You could do something like the `yj` command I shared with you on Slack to apply a filter to buildpack.toml prior to running and create-package and that might help. I'm hoping to get #326 wrapped up in the next week or so. Then we can cut a release and get this shipped.

braunsonm:
From your description on the issue it looks like it still wouldn't work unless you manually move the dependencies to the right directory as well.
I think we'll wait until that issue is also fixed before adopting this... for now we are going to have to keep pinning to old versions.

PR #345, opened 2024-06-05
Automatically add `deprecation_date` field to buildpack dependencies if requested

pbusko:
## Summary

Automatically populate `deprecation_date` for dependencies

## Use Cases

This change adds the `--eol-id` flag to the `update-buildpack-dependencies` cmd. When the flag is set, the product the EOL date is looked up for a product (the `--eol-id` flag), formatted and added to the `deprecation_date` field.

## Checklist


dmikusa:
I don't feel like we should pull in a dependency on github.com/kobayashi/eol. It's a 0.1.0 release and hasn't been updated in over a year (last commit). It doesn't show strong signs of life. If I'm missing something there please let me know.

It may be a little more work, but there are only three API calls for the API itself, https://endoflife.date/docs/api. Do you have any objections to just making HTTP calls directly to the API? You could add a client under the `internal` package, so we don't need to worry about exporting that as public, it'll be just for the library.

pbusko:
> It's a 0.1.0 release and hasn't been updated in over a year (last commit)

good point, I've moved the implementation to the internal package

c0d1ngm0nk3y (approved):
Some smaller comments...

c0d1ngm0nk3y:
@dmikusa We call the API directly instead of using the additional dependency like you proposed in [this comment.](https://github.com/paketo-buildpacks/libpak/pull/345#issuecomment-2153014136) Maybe you can have another look.

pbusko:
@dmikusa is there anything else blocking the PR from being merged?

dmikusa (approved):

PR #366, opened 2024-09-23
Add skip-path Feature for Dependency Mirrors

bitgully:
## Summary
Adds `skip-path` feature for dependency mirrors.
This provides the option of skipping path segments of the original location as discussed [here](https://github.com/orgs/paketo-buildpacks/discussions/298) in regards to the `/maven2` issue.

Usage: Can be used by a comma separated argument as in https://mirror.example.org,skip-path=/path-to-skip

## Use Cases
1) Download all dependencies from https://mirror.example.org but skip `/maven2` in the mirror's path if this is part of any original location.
BP_DEPENDENCY_MIRROR=https://mirror.example.org,skip-path=/maven2

2) Download all dependencies from https://mirror.example.org using the original path but skip `/maven2` on the mirror if the original path at https://repo1.maven.org includes this segment.
BP_DEPENDENCY_MIRROR=https://mirror.example.org
BP_DEPENDENCY_MIRROR_REPO1_MAVEN_ORG=https://mirror.example.org,skip-path=/maven2

This feature can also be used in `dependency-mirror` bindings.

## Checklist


dependency_cache.go
```diff
@@ -454,16 +454,52 @@ func (DependencyCache) verify(path string, expected string) error {
...
+// Parses a raw mirror string into a map of arguments.
+func parseMirror(mirror string) map[string]string {
+
+	mirrorArgs := map[string]string{
+		"mirror":    mirror,
+		"skip-path": "",
+	}
+
+	// Split mirror string at commas and extract specified arguments.
+	for _, arg := range strings.SplitN(mirror, ",", -1) {
```
dmikusa, line 481:
```suggestion
	for _, arg := range strings.Split(mirror, ",") {
```

It's probably simpler to just split (I had to look at the docs to see what -1 means). It is the same behavior as this.

> It is equivalent to [SplitN](https://pkg.go.dev/strings#SplitN) with a count of -1.

Unless we want to put a limit on things. You could probably `SplitN(mirror, ",", 2)` if we wanted to lock this down a bit more. You should really only have at most one `,` cause we only support one comma separated argument at this time and you either have that argument or you have nothing.

dependency_cache_test.go
```diff
@@ -440,6 +440,50 @@ func testDependencyCache(t *testing.T, context spec.G, it spec.S) {
...
+					ghttp.RespondWith(http.StatusOK, "test-fixture"),
+				))
+
+				dependencyCache.DependencyMirrors["127.0.0.1"] = mirrorUrl.Scheme + "://user%3Apa%24%24word%2C%40" + mirrorUrl.Host + "%2Fescaped"
+				a, err := dependencyCache.Artifact(dependency)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(io.ReadAll(a)).To(Equal([]byte("test-fixture")))
+			})
+
```
dmikusa, line 470: Can you add a test that covers the case where the mirror & skip-path need to be PathUnescaped to work correctly?

dmikusa (changes requested):
Thanks so much for this PR & sorry for the delay. It's looking good, just two minor things for your consideration.

dependency_cache.go
bitgully: The simpler split function might be more handy, compared to limiting the number of arguments. Should we have additional ones in the future, we don't need to touch this part of the code again.

I'd consider the "mirror=" key for the URL as optional and added a test case to cover this format.

dmikusa (approved):

PR #395, opened 2025-03-07
Support pre-package commands with arguments

### LINKED ISSUE #856: Local `create-package` broken by multiarch
When calling `create-package --destination target --version 0.0.1`, it fails with

```
Paketo Buildpack for NPM Install 0.0.1
  https://github.com/paketo-buildpacks/npm-install
  Creating package in target
  Pre-package with ./scripts/build.sh --target linux/amd64 --target linux/arm64
unable to execute pre-package script ./scripts/build.sh --target linux/amd64 --target linux/arm64
unable to start PTY
fork/exec ./scripts/build.sh --target linux/amd64 --target linux/arm64: no such file or directory
```

As far as I can tell, this problem was introduced with https://github.com/paketo-buildpacks/npm-install/pull/806

## Expected Behavior

A local package is created (as before) which can be used for local testing (calling `pack buildpack package --target linux/arm64` afterwards)

## Current Behavior

It will try to execute the file `./scripts/build.sh --target linux/amd64 --target linux/arm64`. So no arguments are allowed to be set to `metadata.pre-package`.

## Possible Solution

`build.sh` could be something simple like [here](https://github.com/paketo-buildpacks/bellsoft-liberica/blob/main/scripts/build.sh) and always create all architectures.

## Steps to Reproduce

1. `create-package --destination target --version 0.0.1`

## Motivations

Local packaging for testing.

c0d1ngm0nk3y:
#fixes https://github.com/paketo-buildpacks/npm-install/issues/856

## Summary

The `create-package` command will invoke the `pre-package` command from the `buildpack.toml`.

## Use Cases

The build script might receive some arguments for multiarch (e.g. [here](https://github.com/paketo-buildpacks/npm-install/blob/46416f79248638ecb5c89f2678b8ced045583a9f/buildpack.toml#L26))

## Checklist


dmikusa:
Thanks for the PR. I'm on the fence about this. It adds some complexity with the shell parsing library. I think the intent is for that to point to a script, though, so if you needed to set args you'd put those in your script. For paketo buildpacks, we have a `scripts/build.sh` script that we use for this.

Can you expand on your situation/use case? Is there a reason you couldn't add a wrapper script that includes your custom arguments? The arguments are going to be static in the buildpack.toml, so I don't think it would be any more flexible to have them in the toml file as opposed to a script. Thanks

c0d1ngm0nk3y:
> It adds some complexity with the shell parsing library.

Can you maybe better explain where you see the complexity. Without using a library, I would totally see your point. But we use the same library as [paketo-buildpacks/procfile](https://github.com/paketo-buildpacks/procfile/blob/4dd149918776b964d049ebdfbc397331080cb84d/procfile/build.go#L51) in the exact same way.

 > I think the intent is for that to point to a script

But being able to call a script only without arguments, sounds kinda like a bug. Doesn't it?

> so if you needed to set args you'd put those in your script. For paketo buildpacks, we have a scripts/build.sh script that we use for this.

That would be an easy workaround for sure. But `npm-install` decided to add arguments to `build.sh`. Which I prefer tbh, because otherwise you would hardcode the targets in 2 places - `buildpack.toml` and `build.sh`

@pacostas Any thoughts on this?

> Can you expand on your situation/use case?

It is all in the linked issue, but basically it is as simple as being able to call `create-package` for the `npm-install` buildpack.

> Is there a reason you couldn't add a wrapper script that includes your custom arguments?

They are not really custom arguments, but the way the `build.sh` works for this buildpack.

> The arguments are going to be static in the buildpack.toml, so I don't think it would be any more flexible to have them in the toml file as opposed to a script.

But what if I want to add a target or (more likely) remove one for local use. I would have to touch 2 places. Right?

c0d1ngm0nk3y:
> Shell parsing is a dangerous task, IMHO. It seems simple on the surface but has a lot of corner cases and risk.

I wouldn't call it dangerous. The worst think that can happen is that it dos not work. But it is not simple for sure. Even if it only works for 80% of the cases, that is 80% more as without.

> I think it's something we should avoid attempting to do if there are other options.

An option could be to shell out and let the shell take care of it. Do you think that might be an option?

> If you can use create-package to build packit based buildpacks, that's great, but it's not a scenario that libpak tests or guarantees.

It is really a coincidence that this buildpack is `packit` based. The use of `create-package` has nothing to do with the library and it worked before. To split the tooling for the community based on the library sounds like an odd choice to me tbh. I always use `create-package` - if I now have to check the used library first, that is odd. Specifically since there is no technical reason.

PR #408, opened 2025-03-24
Allow layer contributors to contribute process types

### LINKED ISSUE #349: Add interface for layer contributor to also contribute process types
## Describe the Enhancement

As a layer contributor, I can't presently set process types. That can only happen in the build func. Enable a layer contributor to set process types, so that logic can move out of build into the layer contributors which often are the ones with the most knowledge about the command to run.

This should be optional though. If the layer contributor does not want to contribute processes, that's OK. If the layer contributor doesn't have enough information to set the process types, that's OK too, and they should still be possible to be set in the build func.

## Motivation

In simple cases, which is where layer contributor excels the layer contributor can contribute a process type and then the logic for this is more cleanly separated in the layer contributor versus in the build func.

dmikusa:
## Summary

Resolves #349

This introduces a new interface `ProcessContributable` which extends `Contributable`. It introduces a new method that must be implemented `ProcessTypes`. This method will be called and allows you to return the list of process types that should be added.

The `ContributableBuildFunc` method is modified to check the type. If it is just `Contributable` then it adds a layer to the BuildResult. If it is `ProcessContributable` then it will add the layer and it will add the list of processes to the BuildResult too.

This should actually be a backward-compatible way to implement this, which means existing Contributable implementations do not need to change. However, we do not have any existing Contributable implementations as this is a new interface in v2. At the same time, it should make migration to v2 easier as this is closer to the v1 implementation.

An alternative implementation for this would be to just have one interface `Contributable` and to add the `ProcessTypes` method to that. This would mean that for implementations that don't care to return process types, they would still need to implement the method. A similar approach would be to modify the `Contribute` method on the `Contributable` interface to pass the process types back from the `Contribute` method. That would mean every implementation needs to deal with process types, but it could just return an empty list which is easy enough. The downside of that is then `Contributable` becomes more multi-purpose. I don't really have a strong preference on how we implement this, but went with the current implementation as it is flexible but it separates concerns. If there's a strong preference for a different implementation, please let me know.

This also removes a couple tests which were for old API versions, and updates mockery calls.


modulo11:
Should this PR really be merged into https://github.com/paketo-buildpacks/libpak/tree/gh_287? How are they connected?

dmikusa:
I just have it queued up that way cause that's the order I was working on them. I'll merge it after that PR.

layer_test.go
```diff
@@ -881,4 +792,64 @@ func testLayer(t *testing.T, context spec.G, it spec.S) {
...
+					mockContributor2,
+				}, nil
+			})
+
+			buildResult, err := buildFunc(libcnb.BuildContext{})
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(buildResult.Layers[0].Name).To(Equal("layer-1"))
+			Expect(buildResult.Layers[1].Name).To(Equal("layer-2"))
+			Expect(mockContributor1.Calls).To(HaveLen(2))
```
c0d1ngm0nk3y, line 819: What is this expected to verify? Probably that `Contribute` is called, right?

sbom/mocks/scanner.go
```diff
@@ -0,0 +1,102 @@
+// Code generated by mockery v2.53.3. DO NOT EDIT.
```
c0d1ngm0nk3y, line 1: Where is this file coming from? Is this related to this pr?

dmikusa: It is auto generated. The `go generate` command can be run to run code generators, which we use to run mockery. What it generates is dictated by comments in the code like [this one](https://github.com/paketo-buildpacks/libpak/pull/408/files#diff-c624765f4df697256be192f9789ec38a15b70661f77956b29d03c81856fa8414R91).

dmikusa: This one surprised me. It looks like someone, probably me, forgot to run `go generate` in the past and we never generated the mock for this Interface.

layer_test.go
dmikusa: This is a little painful to test. The ContributableBuildFunc takes in the mock Contributables which in turn get cycled through. This is trying to verify that we ran them and that it used the name in the expected way, which is to set the name on the layer produced by ContributableBuildFunc. The HaveLen check is to make sure that it called both Name and Contribute.

Probably other ways to test this. If there's a preference for doing it a different way, in particular one that's more clear, I'm fine with that. I have no particular attachment to this approach.

anthonydahanne (approved):

PR #412, opened 2025-04-01
Use toml encoder/decoder for buildpack metadata unmarshalling

pbusko:
## Summary

Instead of manual traversing through the map, the `toml.Encode` and `toml.Decode` can be used

## Use Cases

The use case is to not re-implement the unmarshaller

## Checklist

### DIFF
--- a/buildmodule.go
+++ b/buildmodule.go
@@ -17,6 +17,7 @@
 package libpak
 
 import (
+	"bytes"
 	"fmt"
 	"net/url"
 	"os"
@@ -27,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/BurntSushi/toml"
 	"github.com/Masterminds/semver/v3"
 	"github.com/heroku/color"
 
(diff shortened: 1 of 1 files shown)


dmikusa:
My initial thought is to say that it was done this way for a reason, but when I took a glance back at the commit logs I don't see one specifically listed. It's not to say there's no reason, but not one that's documented, unfortunately.

I did double check the test coverage, and it looks reasonable, so given that tests are not breaking that's giving me some confidence. I'm just having a hard time getting past that "why was it done this way originally?" question. 😄 

@anthonydahanne What do you think about this one?

pivotal-david-osullivan:
Will this affect the `deprecation_date` parsing? This is done in the 'manual' code and is relevant when comparing layer data for the restore/reuse of the cached stuff.

The build + rebuild scenario should probably be tested specifically with this change to make sure it doesn't break caching

pbusko:
> Will this affect the `deprecation_date` parsing? This is done in the 'manual' code and is relevant when comparing layer data for the restore/reuse of the cached stuff.
> 
> The build + rebuild scenario should probably be tested specifically with this change to make sure it doesn't break caching

Encoder uses the same  RFC 3339 for encoding `time.Time` fields, while decoder supports some additional formats.

anthonydahanne (approved):
while I can't talk about the past / history on this one, it looks like it's way simpler and robust; if the tests passed as well, and they cover enough ground, I'm all in!

dmikusa (approved):

Issue #327, opened 2024-04-10
Implement RFC 0059 - Update metadata to be consistent with RFC standard


dmikusa:
## Describe the Enhancement

In [RFC 0059](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0059-standard-dependency-metadata-format.md) we introduce a standard metadata format. We need to update libpak to support this format.

## Possible Solution

Presently, most of the libpak metadata looks like:

```
  [[metadata.dependencies]]
    cpes = ["cpe:2.3:a:oracle:jdk:1.8.0:update402:*:*:*:*:*:*:*"]
    id = "jdk"
    name = "BellSoft Liberica JDK"
    purl = "pkg:generic/bellsoft-jdk@8.0.402?arch=amd64"
    sha256 = "e8239e9e5f046a460b3c2b92019714f9ce0da96b272c1fa78016d903cb547c94"
    source = "https://github.com/bell-sw/Liberica/releases/download/8u402+7/bellsoft-jdk8u402+7-src.tar.gz"
    source-sha256 = "da4acb30a3d613fc14766f11ba2159aaddf0428e5adda0b4bed75465f910d426"
    stacks = ["io.buildpacks.stacks.bionic", "io.paketo.stacks.tiny", "*"]
    uri = "https://github.com/bell-sw/Liberica/releases/download/8u402+7/bellsoft-jdk8u402+7-linux-amd64.tar.gz"
    version = "8.0.402"

    [[metadata.dependencies.licenses]]
      type = "GPL-2.0 WITH Classpath-exception-2.0"
      uri = "https://openjdk.java.net/legal/gplv2+ce.html"
```

The new format would look like:

```
  [[metadata.dependencies]]
    checksum = "<dependency algo:checksum>"
    id = "<dependency ID>"
    uri = "<dependency URI>"
    version = "dependency version"

    arch = "<dependency compatible architecture>" #optional
    cpes = [ "<dependency cpe>" ] #optional
    eol-date = "<dependency eol>" #optional
    name = "<dependency name>" #optional
    os = "<dependency compatible OS>" #optional
    purls = [ "<dependency purl>" ] #optional
    source = "<dependency source URI>" #optional
    source-checksum = "<dependency source algo:checksum>" #optional
    strip-components = <number of directories to strip off dependency artifact> #optional

    [[metadata.dependencies.distros]] #optional
      name = "<compatible OS distribution name>"
      version = "<compatible OS distribution version>" #optional

[...]

PR #414, opened 2025-04-03
Implementation for the RFC0059

pbusko:
## Summary

Part of https://github.com/paketo-buildpacks/libpak/issues/327

## Use Cases

Backward-compatible Implementation of the [RFC0059](https://github.com/paketo-buildpacks/rfcs/blob/main/text/0059-standard-dependency-metadata-format.md)

## Checklist

### DIFF
--- a/layer.go
+++ b/layer.go
@@ -24,12 +24,9 @@ import (
 
 	"github.com/BurntSushi/toml"
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/heroku/color"
 
 	"github.com/buildpacks/libcnb/v2"
-
-	"maps"
 
 	"github.com/paketo-buildpacks/libpak/v2/log"
 	"github.com/paketo-buildpacks/libpak/v2/sbom"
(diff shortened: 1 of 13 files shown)


layer.go
```diff
@@ -188,65 +190,27 @@ func (l *LayerContributor) checkIfMetadataMatches(layer libcnb.Layer) (map[strin
...
-		return false, fmt.Errorf("%w (expected layer)", err)
-	}
-
-	if err := l.normalizeDependencyDeprecationDate(layerM); err != nil {
+	if err := l.removeDependencyDeprecationDate(layerM); err != nil {
 		return false, fmt.Errorf("%w (actual layer)", err)
 	}
 
-	return reflect.DeepEqual(expectedM, layerM), nil
+	return cmp.Equal(expectedM, layerM, cmpopts.EquateEmpty()), nil
```
dmikusa, line 197: Can you expand on the changes in this file?

1. It looks like `cmp.Equal` is going to help with comparisons. Does this handle some of the issues that we have seen previously, like empty vs nil and the time matching? I ask because this is a really delicate bit of code here that's been very problematic. When it fails, it's not always obvious to users that it's failing, but it results in cache fails and much slower builds. I just want to be very careful with any changes here.

2. I think we also need to consider if and how we want to handle the case where someone has done a build with a v1 buildpack and is now building with v2. This is going to cause the shape of all of the dependencies to be different than what's in the layer metadata. That probably just means a bunch of cache misses and rebuilds, which isn't the worst thing. I actually worry more about the opposite, where the cache doesn't miss, and there is no rebuild when we should rebuild because there are major changes here (like maybe with helpers?). This may be a separate issue, but I'm wondering if we want to put the libpak major version in the layer metadata. That way, we guarantee a rebuild of all layers when the libpak version changes. If you think this should be a separate issue, feel free to open a new issue for this.

3. I'm also curious about removing the dependency depreciation dates. My brain might not be working the best right now, but I'm not following why we'd want to remove these. It would be nice if we could get back to libpak seeing the metadata comparison as a generic thing, where we are not having to dig into the metadata and magically change certain parts of it. Maybe the answer is that it's just to complicated to do generically, and if that's the answer then maybe we should look at making some types for this deserializing to those types instead of using the map/cast approach? The RFC didn't standardize the "configuration"/"dependency" key (probably should have), but I think that "configuration" and "dependencies" are consistently used, so we could have a type for that too.

dmikusa (commented):
Looking good, some thoughts...

layer.go
pbusko: > It looks like cmp.Equal is going to help with comparisons. Does this handle some of the issues that we have seen previously, like empty vs nil and the time matching?

The only issues that was manually handled here were the `nil` slices and `time.Time` comparison. Both cases are handled correctly by the `cmp.Equal`

> I think we also need to consider if and how we want to handle the case where someone has done a build with a v1 buildpack and is now building with v2.

Technically the v2 version will result in a 100% cache misses, because it completely changes how the layer metadata is generated from the dependency object. If we want to preserve cache hits, we have to manually identify the metadata structure from v1 and compare only the fields that are used by v2

> I'm also curious about removing the dependency depreciation dates. 

There's nothing to compare them to, because v2 changed the fields for the layer metadata, but that is true for other fields as well though

dmikusa:
I think this code looks good. I'm going to try and do some testing with it soon.

@paketo-buildpacks/java-maintainers Any thoughts?

dmikusa:
I think we can probably merge this. I haven't gotten to test it thoroughly yet because I need some v2 buildpacks to test it with. I think if we merge this, we can unblock cutting a release candidate of libpak v2, which will let us move forward updating buildpacks & actually testing this more.

If anyone has any objections, let me know. Otherwise, I'll merge this tomorrow (Wed). Thanks!

dmikusa:
@pbusko Can you update your PR one more time? Thanks!

dmikusa (approved):

PR #428, opened 2025-07-24
Add binding predicate OfKeyWithValueFromSecret

tomkennedy513:
## Summary
This predicate allows finding a binding that has a secret with a field containing a particular value

## Use Cases
When dealing with bindings that may have come from parsing VCAP_SERVICES, the type field in the binding may be "user-provided" which doesn't match an actual binding type. Instead, we can look for a type field in the secret to find the true type, allowing users to override a binding type when creating a user-provided service in Cloud Foundry

## Checklist

### DIFF
--- a/bindings/resolve.go
+++ b/bindings/resolve.go
@@ -34,6 +34,19 @@ func OfType(t string) Predicate {
 	}
 }
 
+// OfKeyWithValueFromSecret returns a Predicate that returns true if a given binding has Key that matches k with Value that matches v.
+// The comparison is case-insensitive.
+func OfKeyWithValueFromSecret(k string, v string) Predicate {
+	return func(bind libcnb.Binding) bool {
+		value, exists := bind.Secret[k]
+		if !exists {
+			return false
+		}
+
+		return strings.EqualFold(value, v)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


dmikusa:
The general idea of filtering by the secret is fine by me. 

I would prefer that the method be more generic though, so it's not tied to vcap_services. One doesn't know what is in Secret, it's a map of opaque key/value strings. Would you be OK with changing this so it's more generic? For example, instead of `OfTypeFromSecret(t string)`, where `type` is something in secret, we could do something like `OfKeyFromSecret(key string, val string)` and if key is in secret and contains val value, then it's true.

I believe that would work for what you're doing, `OfKeyFromSecret("type", "foo")`.

tomkennedy513:
> The general idea of filtering by the secret is fine by me.
> 
> I would prefer that the method be more generic though, so it's not tied to vcap_services. One doesn't know what is in Secret, it's a map of opaque key/value strings. Would you be OK with changing this so it's more generic? For example, instead of `OfTypeFromSecret(t string)`, where `type` is something in secret, we could do something like `OfKeyFromSecret(key string, val string)` and if key is in secret and contains val value, then it's true.
> 
> I believe that would work for what you're doing, `OfKeyFromSecret("type", "foo")`.

I like that a lot, I will make the change

dmikusa (approved):

